PR #2, opened 2018-12-20
EU ecl card

nimek2:
https://v2--europa-component-library.netlify.com/eu/components/card/usage/


nimek2:
closing since locally we have improved version will be relested after approval of PR 12

PR #14, opened 2019-02-07
feat: EU component Button

nimek2:
We did all as in previously accepted PR. Soo take a look please


src/eu/packages/eu-component-button/button.story.js
```diff
@@ -0,0 +1,61 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, text } from '@storybook/addon-knobs';
```
planctus: So, this is a common pattern, i see. We have two systems, EC and EU and for how "similar" they might look, they are independent and we should avoid confusion.
Here we are only dealing with a demo of the component, in the end, the twig component per se is not affected by this, but still we want to keep things separated, so please use only resources from the right ecosystem, in this case EU ;)

hapertown: Fixed in latest commit [4a6eb06](https://github.com/ec-europa/ecl-twig/pull/14/commits/4a6eb062c24d03bb9dc0a220fe8ec66ae6f01d4d)

hapertown:
I changed resources on the EU but the tests come out incorrectly. There are no modules whose names I have changed (see details in logs)

hapertown:
@planctus see log from tests https://drone.fpfis.eu/ec-europa/ecl-twig/56/5. We miss one/two packages from the EU environment (eg. @ecl/eu-specs-button/demo/data--primary). This error appeared after the change sources from EC to EU. You should not add these modules first to enviroment?

emeryro:
Concerning the missing files, you miss the dependency in your eu button package (it is there in EC, that's why it worked):
```
"devDependencies": {
    "@ecl/eu-specs-button": "2.0.0"
},
```

hapertown:
> Concerning the missing files, you miss the dependency in your EU button package (it is there in EC, that's why it worked):
> 
> ```
> "devDependencies": {
>     "@ecl/eu-specs-button": "2.0.0"
> },
> ```

Thanks for the hint, I fixed it in the last commit. [4ce88f4](https://github.com/ec-europa/ecl-twig/pull/14/commits/4ce88f4379ede84cc8cb956c0cdb52f1b1a82e7a)

hapertown:
We already have merged icons, so I add support for this component and we can finally merge the component.

planctus:
I put the @wip tag just in order to make it evident that we are waiting for the icons support. ;)

hapertown:
@planctus @emeryro 
I commited last changes - generally EU Icon support and completed missing docs files.
I have two questions:
1. See: [test](https://github.com/ec-europa/ecl-twig/blob/fb969c66a3f94415b043a1e7960c5eb0112ab1bb/src/eu/packages/eu-component-button/button.test.js) for CTA Button.  I have rewritten some parameters from the demo file to fit the structure into EU Icon. Should not we change the data format in the `@ecl/eu-specs-button` package to match the current EU Icon structure?

2. I have a similar problem with Storybook icons like the one I described in https://github.com/ec-europa/ecl-twig/pull/16#issuecomment-463175695. When we can solve this problem in https://github.com/ec-europa/ecl-twig/pull/16, I will be able to improve it also here.

planctus:
So, @hapertown, [here](https://github.com/ec-europa/ecl-twig/pull/16) maybe you have a solution for the icons select.
I still have to better understand the first question, but i see that we miss the "disabled" state for the buttons, we have it in ECL v2.

hapertown:
> So, @hapertown, [here](https://github.com/ec-europa/ecl-twig/pull/16) maybe you have a solution for the icons select.
> I still have to better understand the first question, but i see that we miss the "disabled" state for the buttons, we have it in ECL v2.

I added my own solution based on your code. I just modified the TWIG file differently, you can check it out.
I saw this option disabled on a demo, but when I implemented it locally, this attribute did not work, so I thought that this attribute is no longer supported. Now is ok. Sorry for my mistake.

hapertown:
> I still have to better understand the first question,

I mean test file, in particular the icon structure. In the `@ecl/eu-specs-button/demo/data-call` package we have test data in the form of:

`
label: 'Call to action button',
icon: {
	shape: 'ui--corner-arrow',
	size: 'xs',
	transform: 'rotate-90',
	},
variant: 'call',
`

but expected structure is

`label: 'Call to action button',
icon: {
	type: 'ui',
	name: 'corner-arrow',
	size: 'xs',
	path: '/path-to-file-icon',
transform: 'rotate-90',
},
variant: 'call',`

that's why in the `button.test.js` file I've overwritten the corresponding values in this way

`dataCall.icon.path = 'static/icons.svg';
dataCall.icon.type = 'ui';
dataCall.icon.name = 'corner-arrow';
`

for the tests to be generated correctly.

My question is about the `@ecl/eu-specs-button/demo/data-call` package - should not we change the test data to those that correspond to our current structure?

emeryro:
Basically, the structure of the icon is different in your implementation, right?
As demo data are just, well, demo data, you can apadt it on your side if needed.

I don't have the full history of discussion, but what is the reason why the icon structure is different?

hapertown:
> Basically, the structure of the icon is different in your implementation, right?
> As demo data are just, well, demo data, you can apadt it on your side if needed.
> 
> I don't have the full history of discussion, but what is the reason why the icon structure is different?

The current structure is due to the fact that when we implemented the EC / EU Icon (or rather when the target structure was created) we did not have ready-made tests or structure proposals from your side.
Therefore, in order for the whole component to be maximally configurable and easy to use, each most important option has been separated into a separate paramter (eg instead of `ui-corner-arrow` we have a separate `ui` and `corner-arrow`).
I think that in our case doesn't matter - for the demo we can always overwrite the values, while the documentation clearly defines what parameters should be used during the production creation of components.

emeryro:
One that thing that should be added to button is the possibility to put icon before or after (like in links).
It's true that is not part of released version of ECL, but we already did it in [dev version](https://v2-dev--europa-component-library.netlify.com/storybook/ec/index.html?selectedKind=Button&selectedStory=primary&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).
That will save time later (no need to update the component again in a couple of weeks)

hapertown:
> One that thing that should be added to button is the possibility to put icon before or after (like in links).
> It's true that is not part of released version of ECL, but we already did it in [dev version](https://v2-dev--europa-component-library.netlify.com/storybook/ec/index.html?selectedKind=Button&selectedStory=primary&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).
> That will save time later (no need to update the component again in a couple of weeks)

Ok, in that case I will add this option in next commit. Give me some time.

hapertown:
> One that thing that should be added to button is the possibility to put icon before or after (like in links).
> It's true that is not part of released version of ECL, but we already did it in [dev version](https://v2-dev--europa-component-library.netlify.com/storybook/ec/index.html?selectedKind=Button&selectedStory=primary&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).
> That will save time later (no need to update the component again in a couple of weeks)

Added in https://github.com/ec-europa/ecl-twig/pull/14/commits/df3533481f78bcae2cf13bc240f0ddc1dcc88bd1

src/eu/packages/eu-component-button/button.html.twig
```diff
@@ -0,0 +1,66 @@
...
+    - "label"
+#}
+
+{# Internal properties #}
+
+{% set _css_class = 'ecl-button' %}
+{% set _extra_attributes = '' %}
+{% set _label = label|default('') %}
+{% set _variant = variant|default('primary') %}
+{% set _type = type|default('button') %}
```
emeryro: default button type should be set to submit

hapertown: Updated on https://github.com/ec-europa/ecl-twig/pull/14/commits/4693f1181179b688854415165e35c950d848ba3e

hapertown: And update tests in https://github.com/ec-europa/ecl-twig/pull/14/commits/45af0a88a12907129f82b9b8b86473f9c52f3815

nimek2:
@planctus @emeryro 

I think all is done here so please merged and we can move forward :)

PR #17, opened 2019-02-08
feat: EC component Tag

nimek2:
Please nogte that this is deoendent on CC icon so have to be merger after that


hapertown:
For me, locally tests come out ok - later I will check what may be the reason for the wrong tests.

hapertown:
@planctus the tests have been fixed in the last commit - you can review component now.

src/ec/packages/ec-component-tag/docs/tag.md
```diff
@@ -0,0 +1,36 @@
...
+## Tag
+
+### Parameters
+
+- "tag" (associative array) (default: 'predefined structure below')
+  - "path" (string) (default: '') - path/url for tag (if needed - only for type 'link')
+  - "type" (string) (default: '') - type of tag (can be 'link', 'button', 'removable')
+  - "label" (string) (default: '') - tag label
+- "default_icon_path" (string ) (default: '') - path for the icon image (need to render Icon component if tag is removable)
+- "extra_classes" (optional) (string) (default: '') Extra classes (space separated) for the icon
```
emeryro: typo: this is tag and not icon here (and next line too)

src/ec/packages/ec-component-tag/tag.html.twig
```diff
@@ -0,0 +1,65 @@
...
+  {% endfor %}
+{% endif %}
+
+{# Print the result #}
+{% if _tag.type!='link' %}
+  <button class="{{ _css_class }}" type="button"{{ _extra_attributes|raw }}>
+    {{ _tag.label }}
+    {% if _tag.type=='removable' %}
+      <span class="ecl-tag__icon">
+        {% include '../ec-component-icon/icon.html.twig' with { icon: { type: 'ui', name: 'close', size: 'xs', path: default_icon_path }, extra_classes: 'ecl-tag__icon-close' } %}
```
emeryro: We recently added the extra attribute `focusable="false"` to all svg icons, to prevent a bug on Internet Explorer. Could you add it on your side too?
See https://v2--europa-component-library.netlify.com/ec/components/icon/code/

src/ec/packages/ec-component-tag/docs/tag.md
hapertown: I don't know what you mean. The variable `default_icon_path` defines here the path to the file with icons, which is required for the tag type `removable`. Do you mean to change the variable name and description or define this path in the structure of the tag object. e.g. 
`tag: { type: 'removable', icon_path: '/pathtoicon' }`?

emeryro: no it's just the comment for extra_classes and extra_attributes. It should be tag instead of icon (but I agree that it is minor)

hapertown: Ok I see my mistake, I will correct everything after merging with the EC Icon.

src/ec/packages/ec-component-tag/tag.html.twig
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/17/commits/069ce0527105026e409f5ea3e42803bc88e2268b

src/ec/packages/ec-component-tag/docs/tag.md
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/17/commits/069ce0527105026e409f5ea3e42803bc88e2268b

nimek2:
All fixed here and test passed please make final review

emeryro:
Component seems good.
Could you just add a devDependency to the "spec" component, theway it has been done [here](https://github.com/ec-europa/ecl-twig/blob/02d605e75f17fa158333819fd8be4fa134bc542c/src/ec/packages/ec-component-icon/package.json#L12)?
In this case it would be `"@ecl/ec-specs-tag": "~2.1.0"`
We consider it as a good practice, to keep the component close to its specification.
Thanks

hapertown:
> Component seems good.
> Could you just add a devDependency to the "spec" component, theway it has been done [here](https://github.com/ec-europa/ecl-twig/blob/02d605e75f17fa158333819fd8be4fa134bc542c/src/ec/packages/ec-component-icon/package.json#L12)?
> In this case it would be `"@ecl/ec-specs-tag": "~2.1.0"`
> We consider it as a good practice, to keep the component close to its specification.
> Thanks

Fixed in latest commit. Should I always add to every component `@ecl/ec[eu]-specs-[component]`?

emeryro:
Yes, please do so for every components

PR #18, opened 2019-02-12
fix: EC icon - Add focusable and aria-hidden parameter

nimek2:


emeryro:
While we are here, could you also add aria-hidden="true" for icons? That way we will be in sync.

hapertown:
Aria hidden added. I will try to fix the tests and we will be able to merge.

hapertown:
> Please fix the error reported by Drone and that will be good

Drone has only 1 error:
`› 1 snapshot obsolete.
   • EC - Icon Branded - icon instagram renders correctly 1`

The snapshot files were generated locally without any problems. Do you have an idea why this test is skipped? I will be grateful for help.

emeryro:
I can reproduce the error locally. Did you run `yarn` and `yarn dist` before running your tests?
(not sure about `yarn dist` actually, but who knows)

hapertown:
> I can reproduce the error locally. Did you run `yarn` and `yarn dist` before running your tests?
> (not sure about `yarn dist` actually, but who knows)

I'll check it.

hapertown:
@emeryro 
Tests fixed in commit [b6fd402](https://github.com/ec-europa/ecl-twig/pull/18/commits/b6fd402d0458a8438a5f4992ea330648627a1a9e) - the component seems ready to merge. Thanks for help.

hapertown:
If everything is ok, please merge it with the master so that I can fix to EC Tag and EC Link asap. Also let me know if there are any errors that we should fix.

PR #20, opened 2019-02-13
feat: component text input

emeryro:
Add component text input (EC and EU)


src/ec/packages/ec-component-text-input/docs/text-input.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Text input
+
+### Parameters:
+
+- "id" (string) (default: '')
+- "disabled" (boolean) (default: false)
+- "invalid" (boolean) (default: false)
+- "invalidIconLabel" (string) (default: '')
```
yhuard: We need to talk about the naming conventions: either we stick to snake_case or to camelCase, but we don't mix both

yhuard: Actually, everywhere else, snake_case is used :) no need to discuss it, let's just update these variables names

src/ec/packages/ec-component-text-input/__snapshots__/text-input.test.js.snap
```diff
@@ -0,0 +1,42 @@
...
+    placeholder=\\"Placeholder\\"
+    class=\\"ecl-text-input\\"
+  />
+</div>
+"
+`;
+
+exports[`EC - Text input Disabled renders correctly 1`] = `
+"<div class=\\"ecl-form-group ecl-form-group--text-input\\">
+  <input
```
yhuard: This is not a disabled input. Please make sure the mapping between `import dataDisabled from '@ecl/ec-specs-text-input/demo/data--disabled'` and what the component consumes matches

src/ec/packages/ec-component-text-input/__snapshots__/text-input.test.js.snap
```diff
@@ -0,0 +1,42 @@
...
+    type=\\"text\\"
+    placeholder=\\"Placeholder\\"
+    class=\\"ecl-text-input\\"
+  />
+</div>
+"
+`;
+
+exports[`EC - Text input With error renders correctly 1`] = `
+"<div class=\\"ecl-form-group ecl-form-group--text-input\\">
```
yhuard: this is not an error input, same remark as above

src/eu/packages/eu-component-button/docs/cta.md
```diff
@@ -14,7 +14,7 @@ npm install --save @ecl-twig/eu-component-button
 - "variant" (string) (default: 'primary'): variant of button (can be 'primary', 'secondary', 'call', 'ghost', 'search')
 - "type" (string) (default: 'submit'): can be the same type as HTML button - 'submit', 'reset', 'button'
 - "disabled" (bool) (default: false): define if button should be disabled (HTML disabled attribute)
-- "icon_position" [optional] (string) (default: after): Define icon position, can be 'before' or 'after'
+- "icon_position" [optional](string) (default: after): Define icon position, can be 'before' or 'after'
```
yhuard: Is it prettier that removed the space?

Shouldn't we alter the syntax in order to prevent making links? `[optional](string)` -> `(string, optional)`

src/ec/packages/ec-component-text-input/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "name": "@ecl-twig/ec-component-text-input",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Text Input",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@ecl/ec-component-form-feedback-message": "^2.1.0",
```
yhuard: For the sake of consistency, the twig component should only rely on the specs. That's what has been done so far. If we decide to depend on the vanilla implementation (which could make sense), then we have to update the other components as well.

What shall we do?

emeryro: this component is using css classes from vanilla components, so I don't see any reason not to put them in dependencies...
but indeed it could have been done for other components too.

src/ec/packages/ec-component-text-input/text-input.test.js
```diff
@@ -0,0 +1,71 @@
...
+  const template = path.resolve(__dirname, './text-input.html.twig');
+
+  describe('Default', () => {
+    test('renders correctly', done => {
+      expect.assertions(1);
+
+      renderTwigFile(
+        template,
+        {
+          label: 'Label',
```
yhuard: why don't you use the data from the specs?

emeryro: it's not the same structure

PR #21, opened 2019-02-15
feat: Ec component card

nimek2:


hapertown:
I added the initial version of EC Card based on EC Link, EC Tag and EC Icon components.

In the [demo](https://v2--europa-component-library.netlify.com/storybook/ec/index.html?selectedKind=Card&selectedStory=card&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)  are 3 variants (`default`, `tile` and `template`). According to the documentation, I made versions for `default` and `tile`, we do not have any instructions for the variant `template`.

My questions:
1. Currently the link to the picture is absolute. Should I use a something resource to import sample graphics to Card (as well as `defaultSprite` variable in EC Icon)?
2. In the Storybook current `Infos` and `Tags` demo are separated by a comma. In our code, each `tag` or `infos` are arrays of the specified sturcture. Do you want Storybook to also dynamically add `Infos` and `Tags`? If so, are you able to help me implement these solutions?

As always - do not pay attention to this moment for tests - we will fix it at the very end.

emeryro:
Hello,

0. the `template`variant is just a demo of how cards looks like when they are displayed side to side on the grid. So basically you just have to display 3 default cards and 3 title cards

1. example images are not part of a package on our side. They are just considered as public files. I guess it is safe enough to have an absolute url using these public files (it should not change soon).

2. we are also using array of objects for `infos` and `tags`. This is the same structure as yours (see [here](https://github.com/ec-europa/europa-component-library/blob/f67ad0416b5abee71261474d0998aa17d6a23f48/src/systems/ec/implementations/react/packages/card/src/Card.jsx#L146)). The comma separated list is the way how Storybook handles arrays (see [here](https://www.npmjs.com/package/@storybook/addon-knobs#array)). So what we do is display a "Storybook array" of tag or info label, then reform the structure we need (with default value for icon of href). You can see it [here](https://github.com/ec-europa/europa-component-library/blob/f67ad0416b5abee71261474d0998aa17d6a23f48/src/systems/ec/implementations/react/packages/card/stories/Index.jsx#L48).
Honnestly, I don't consider it as a blocker.

hapertown:
I added to Storybook the possibility of adding several elements separated by commas. Thanks for your hints.

Since `template` is just a template for displaying X elements in a row, do we have to implement this to our Storybook? I would then have to make a new TWIG file to implement the grid structure and correctly display the EC Card in a row, which we did not have in the documentation.

src/ec/packages/ec-component-card/card.story.js
```diff
@@ -0,0 +1,112 @@
...
+          },
+          path: defaultSprite,
+        },
+      }));
+
+      return card({
+        card: {
+          title: {
+            type: 'standalone',
+            path: text('Title path', '/example'),
```
emeryro: there is no real point in letting the user change the href of the link (no visual impact).
So I'd just put as a simple string

src/ec/packages/ec-component-card/card.story.js
```diff
@@ -0,0 +1,112 @@
...
+              return 'calendar';
+            }
+
+            if (key === 1) {
+              return 'location';
+            }
+
+            return 'faq';
+          },
+          path: defaultSprite,
```
emeryro: open question here: why don't you use 'static/icons.svg' instead of loading the icons from ECL?
as far as I know it would work too, and that would remove an extra devDependency

src/ec/packages/ec-component-card/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ecl-twig/ec-component-card",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Card",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
```
emeryro: please add a devDependency to ECL vanilla component ("@ecl/ec-component-card": "^2.1.1")

src/ec/packages/ec-component-card/card.story.js
hapertown: In all previous components, we always gave URLs to icons from a specific resource, so that you can easily change URLs easily and globally - I think so. We did it for the first time, probably in EC Icon, and in this way I'm duplicating it.
Do you want to set a specific URL to the icons in this case?

src/ec/packages/ec-component-card/package.json
hapertown: Added in https://github.com/ec-europa/ecl-twig/pull/21/commits/78442bbb33e4a3550799ccb38601a6fc25482e3b

src/ec/packages/ec-component-card/card.story.js
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/21/commits/78442bbb33e4a3550799ccb38601a6fc25482e3b

src/ec/packages/ec-component-card/package.json
hapertown: I updated the package.json file in the ec-components folder. I entered all current modules, can it remain that way? Should I have only the current module in dependencies? In the next components, I just have to enter the next - current - component over which I work?

hapertown:
@emeryro 
I still can not deal with these dependencies in the EC Card
https://drone.fpfis.eu/ec-europa/ecl-twig/149/5

Could you help me with this?

hapertown:
@emeryro could you help me solve file conflicts that block merging branch?

src/ec/packages/ec-component-card/card.story.js
emeryro: it's fine as it is

src/ec/packages/ec-component-card/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@ecl-twig/ec-component-card",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Card",
+  "publishConfig": {
+    "access": "public"
+  },
```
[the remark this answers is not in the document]
hapertown: The same error like here: https://github.com/ec-europa/ecl-twig/pull/28#issuecomment-465988668
Package not found.

src/ec/packages/ec-component-card/docs/card.md
```diff
@@ -0,0 +1,104 @@
...
+    - "path" (string) (default: ''): Link url (href attribute)
+- "extra_classes" (optional) (string) (default: '') Extra classes (space separated) for the Card
+- "extra_attributes" (optional) (array) (default: []) Extra attributes for Card
+  - "name" (string) Attribute name, eg. 'data-test'
+  - "value" (string) Attribute value, eg: 'data-test-1'
+
+### Example for default Card:
+
+```twig
+{% include 'path/to/card.html.twig' with {
```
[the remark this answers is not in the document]
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/21/commits/e08d89f47d4a42379067abda6b8563ec5a0610e3

src/ec/packages/ec-component-card/package.json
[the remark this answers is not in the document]
hapertown: Added in https://github.com/ec-europa/ecl-twig/pull/21/commits/94447b37b953a5a71a6e892e624920c0e0746392

emeryro:
one last thing: make sure to add the card component in `ec-components/package.json`.
then we can merge

hapertown:
> one last thing: make sure to add the card component in `ec-components/package.json`.
> then we can merge

Fixed in https://github.com/ec-europa/ecl-twig/pull/21/commits/8f3f8faa71446ce7ffa98f270937ef80145ff70c

PR #22, opened 2019-02-15
fix: EC Button

nimek2:


hapertown:
I copied the EU Button component and adapted it to the EC documentation. Please, check if everything is okay and you can merge.

PR #23, opened 2019-02-19
feat: EC footer

emeryro:
Add corporate and custom EC footer.

I also had to fix icons in links (icon size were hardcoded).


src/ec/packages/ec-component-link/link.html.twig
```diff
@@ -62,10 +63,10 @@
 {# Print the result #}
 <a href="{{ _link.path }}" class="{{ _css_class }}"{{ _extra_attributes|raw }}>
   {% if _icon.name is not empty and _link.icon_position == 'before' %}
-    {% include '../ec-component-icon/icon.html.twig' with { icon: icon|merge({ size: 'fluid' }), extra_classes: 'ecl-link__icon' } %} &nbsp;
+    {% include '../ec-component-icon/icon.html.twig' with icon|merge({extra_classes: 'ecl-link__icon'}) %} &nbsp;
```
degliwe: this &nbsp; thing should be fixed at the ECL level ecl-link__icon being the right place to add this spacing.

PR #31, opened 2019-02-21
feat: EU component Link

nimek2:


src/eu/packages/eu-component-link/link.story.js
```diff
@@ -0,0 +1,85 @@
...
+          type: 'default',
+          label: text('Label', 'Default link'),
+          path: '/example#link-default',
+          icon_position: iconPosition,
+        },
+        icon: {
+          type: 'ui',
+          name: iconsListSelect,
+          path: defaultSprite,
+        },
```
emeryro: we should set the icon size to `fluid` here, it is currently too big
the same for the other icon below

src/eu/packages/eu-component-link/docs/link.md
```diff
@@ -0,0 +1,42 @@
...
+  - "path" (string) (default: '') - Link url (href attribute)
+  - "icon_position" (string) (default: 'after') - Position of link icon (can be 'before' or 'after') if icon is available
+- "icon" [optional] (associative array) default: A predefined structure such as in the Icon component. All parameters can be freely set, except for the 'size' parameter, which is set permanently as 'fluid'.
+- "extra_classes" (optional) (string) (default: '') Extra classes (space separated) for the icon
+- "extra_attributes" (optional) (array) (default: []) Extra attributes for icon
+  - "name" (string) Attribute name, eg. 'data-test'
+  - "value" (string) Attribute value, eg: 'data-test-1'
+
+### Example:
+
```
emeryro: Please add `<!-- prettier-ignore -->` to prevent extra formatting to be removed (see discussion on slack)

src/eu/packages/eu-component-link/link.story.js
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/31/commits/45a6305895c369887c37648d7ec8c93b36c7d32b

src/eu/packages/eu-component-link/docs/link.md
hapertown: Fixed in https://github.com/ec-europa/ecl-twig/pull/31/commits/45a6305895c369887c37648d7ec8c93b36c7d32b

hapertown:
I have one more question - the current documentation says that the element `<span class="ecl-link__label">` only appears if the link has an icon. Has the documentation changed? From what I remember, this element has always occurred before. Can it stay that way or do you want me to introduce the correct patch right away?

And question # 2 - some other components (such as Site Header) could use the Link component if it would handle the label attribute not as `clean text` but as an `HTML block`. The documentation defines this element as `clean text` (eg. `<a href="">clean text</a>`). Think about whether it is necessary to convert this component into HTML support to use it in other components (eg Site Header), for example: `<a href...> HTML code </a>`

emeryro:
Question 1:
Actually that's a good remark. Indeed on ECL the html structure is as simple as possible if there is no icon (see https://v2-dev--europa-component-library.netlify.com/ec/components/navigation/link/code/).
So it should be the same here. Please update EC and EU while we're at it.
Thanks

Question 2:
On ECL we decided that the link "component" should stay simple, to fulfil the (simple) purpose it has.
If a link require more complex markup (like in the site header), we simply don't use the link component, but use a standard `<a>` with adapted styling. That's what we did on [site header](https://github.com/ec-europa/europa-component-library/blob/70a3990f1dddafbacad31a2b9605f35d1017484c/src/systems/ec/implementations/react/packages/site-header/src/SiteHeader.jsx#L51)

hapertown:
> Question 1:
> Actually that's a good remark. Indeed on ECL the html structure is as simple as possible if there is no icon (see https://v2-dev--europa-component-library.netlify.com/ec/components/navigation/link/code/).
> So it should be the same here. Please update EC and EU while we're at it.
> Thanks
Ok fixed in https://github.com/ec-europa/ecl-twig/pull/31/commits/9ae8d243053656a39868a547220017a7346c14d6, check it please and merge asap. 
We will then be able to add a component EU Card.

emeryro:
> Please update EC and EU while we're at it.

Do you intend to update the EC link component in another PR?

Apart of that EU links are good.
I'll check, but I think that we should focus on EC components first, and add EU version later.

hapertown:
> > Please update EC and EU while we're at it.
> 
> Do you intend to update the EC link component in another PR?

Yes, here: https://github.com/ec-europa/ecl-twig/pull/34. I'm still working on the correct tests and we can merge it.

PR #32, opened 2019-02-22
chore(tests): snapshot DOM nodes instead of strings

yhuard:


src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
```diff
@@ -1,87 +1,147 @@
...
 exports[`EC - Button CTA button - icon after renders correctly 1`] = `
-"<button class=\\"ecl-button ecl-button--call\\" type=\\"submit\\">
-  <span class=\\"ecl-button__container\\">
-    <span class=\\"ecl-button__label\\">CTA Button with icon</span>
-
+<button
+  class="ecl-button ecl-button--call"
+  type="submit"
+>
+  <span
```
yhuard: this is how jest serializes DOM nodes in snapshots

the output is indented only for the purpose to bring more clarity and make it easier to compare snapshots

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
```diff
@@ -1,87 +1,147 @@
...
+    <span
+      class="ecl-button__label"
+    >
+      CTA Button with icon
+    </span>
     <svg
-      class=\\"ecl-icon ecl-icon--fluid ecl-button__icon ecl-button__icon--after\\"
-      focusable=\\"false\\"
-      aria-hidden=\\"true\\"
+      aria-hidden="true"
```
yhuard: attributes are ordered alphabetically

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
```diff
@@ -1,87 +1,147 @@
...
+>
+  <span
+    class="ecl-button__container"
+  >
     <svg
-      class=\\"ecl-icon ecl-icon--fluid ecl-icon--rotate-90 ecl-button__icon ecl-button__icon--before\\"
-      focusable=\\"false\\"
-      aria-hidden=\\"true\\"
+      aria-hidden="true"
+      class="ecl-icon ecl-icon--fluid ecl-icon--rotate-90 ecl-button__icon ecl-button__icon--before"
```
yhuard: if we want to compare the snapshots with a reference — coming from the specs —, then we have to take care of the order of the CSS classes 🤔

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
```diff
@@ -1,87 +1,147 @@
...
+  class="ecl-button ecl-button--call"
+  type="submit"
+>
+  <span
+    class="ecl-button__container"
+  >
+    <span
+      class="ecl-button__label"
+    >
+      Call to action button
```
yhuard: this can be tricky... if the output of the twig template was `<span class="ecl-button__label"> Call to action button </span>`, then the snapshot would add a space before and after `Call to action`. It's not easy to catch it

utils/test-utils/index.js
```diff
@@ -6,6 +6,19 @@ const renderTwigFile = (file, options, cb) =>
...
+      const div = document.createElement('div');
+      div.innerHTML = html.trim();
+
+      return resolve(div.firstChild);
+    })
+  );
+
 module.exports = {
   renderTwigFile,
+  renderTwigFileAsNode,
```
yhuard: for now, we expose the 2 methods but at some point, we should probably harmonize our tests and only use one

unless we want to test the ouput of the twig template as well. but then, should we still run it through prettier?

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
kalinchernev: Looks ok to me, an attribute per line adds clarity. The closing brackets on new lines are like prettier, I like this display.

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
kalinchernev: That's nice, easier to read and check.

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
kalinchernev: Question is: is it important enough to enforce order of classes?

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
kalinchernev: Actually: could it be that's it is prettier changing that in the commit hook?

yhuard: prettier doesn't run on *.snap files 😉

src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
yhuard: the order of the CSS classes in HTML does **not** matter, and that's the issue here: if the specs say `class="ecl-icon ecl-icon--fluid ecl-icon--rotate-90"` and you have `
class="ecl-icon ecl-icon--rotate-90 ecl-icon--fluid`, jest will think there's an error...

utils/test-utils/index.js
kalinchernev: My opinion: `renderTwigFileAsNode` achieves what is necessary.
For what it's worth as feedback from the runner, here's what I get when a difference appears:

```
 FAIL  src/ec/examples/composition/composition.test.js
  EC - Composition
    ✕ renders correctly (57ms)

  ● EC - Composition › renders correctly

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "EC - Composition renders correctly 1".

    - Snapshot
    + Received

    @@ -3,11 +3,11 @@
          ><span class="ecl-button__label">Example button</span></span
        >
      </button>
      
      <blockquote class="ecl-blockquote">
    -   <p class="ecl-blockquote__body">Example body</p>
    +   <span class="ecl-blockquote__body">Example body</span>
        <footer class="ecl-blockquote__attribution">
          <cite class="ecl-blockquote__author">Author name</cite>
        </footer>
      </blockquote>
      "
```

It's as good as it could be, nothing is lost with comparison to the other method. Snapshots are also more readable.

yhuard: and as expected, when you change the order of CSS classes (which doesn't affect the end result):

```diff
          type="submit"
        >
          <span class="ecl-button__container"
            ><span class="ecl-button__label">Search</span
            ><svg
-             class="ecl-icon ecl-icon--fluid ecl-button__icon ecl-button__icon--after"
+             class="ecl-icon ecl-icon--fluid ecl-button__icon--after ecl-button__icon"
              focusable="false"
              aria-hidden="true"
            >
              <use xlink:href="static/icons.svg#general--search"></use></svg
          ></span>
```

😢

kalinchernev: Yes, that is the case with the `renderTwigFile` as well. Yesterday, when I was adding checks on which classes to add and which not, the order was caught again as an error from jest. It's not a problem, I think, as long as the person knows the `u` in `watchAll` ...

PR #34, opened 2019-02-22
fix: Update EC Link structure (twig file and tests)

nimek2:


hapertown:
@emeryro 
I have also allowed myself to update the EC Card test file here https://github.com/ec-europa/ecl-twig/pull/34/commits/313196a45b9ed6ae06eb22c996e50bac97881319 (now the tests are different because the EC Link structure has changed). Can we leave a commit here or should we do it in a separate branch?
Please let me know then I will correctly update the tests file

emeryro:
it's ok to update the cards here, as they rely on the links

hapertown:
@emeryro 
Ok, my tests are ok. Now the bugs in the EC Footer component - could you fix tests, because we did not do this component and I would not like to broke something?

PR #35, opened 2019-02-26
feat: EC component text area

emeryro:


src/ec/packages/ec-component-text-area/text-area.html.twig
```diff
@@ -0,0 +1,83 @@
...
+{% set _placeholder = placeholder|default('') %}
+{% set _rows = rows|default(4) %}
+
+{% set _helper_text = helper_text|default('') %}
+{% set _invalid_text = invalid_text|default('') %}
+{% set _label = label|default('') %}
+
+{% set _css_class = ['ecl-text-area'] %}
+{% if _invalid %}
+  {% set _css_class=_css_class|merge(['ecl-text-area--invalid']) %}
```
kalinchernev: Not a blocker, yet it "hurts" the eyes not having spaces between assignments. Later in the same document, there are spaces. Too pity there is no prettier for that ...

src/ec/packages/ec-component-text-area/text-area.html.twig
```diff
@@ -0,0 +1,83 @@
...
+  {%- endif -%}
+
+  <textarea
+    id="{{- _id -}}"
+    name="{{- _name -}}"
+    rows="{{- _rows -}}"
+    placeholder="{{- _placeholder -}}"
+    class="{{- _css_class -}}"
+    {{- _disabled ? "disabled" -}}
+    {{- _extra_attributes|raw -}}
```
kalinchernev: Quickly checked a very-old implementation. It has `aria-multiline` and `role="textbox"` attributes. Seems like they could be useful for [assisting technologies](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role), would it make sense to include them by default?

emeryro: Good point... I don't know if they have been remove on purpose in ECL 2.
I'll check

emeryro: As it seems, the `role="textbox"` should not be used in textarea (first paragraph [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role), and also https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#Technical_summary)

In a similar way, `aria-multiline` seems to be used only to mimic a textarea, but not in actual textareas.
So I'd leave the implementation as it is.

PR #43, opened 2019-03-01
chore: decide what to do with whitespaces

yhuard:
Let's discuss the whitespace topic here.

- Can you find other problematic solutions?
- Are there other examples you'd want to see in the document? 

`src/ec/packages/ec-component-select/select.html.twig` has been updated in order to reflect the changes discussed here.


emeryro:
to me this is just adding complexity to something that should be simple.
if I take the following code:
```
<div class="ecl-feedback-message">{{ _invalid_text }}</div>
```
if later we add another class, the code has to change:
```
<div class="ecl-feedback-message ecl-feedback-message--something">
  {{- _invalid_text -}}
</div>
```
To me there is a high change that the dash is forgotten here.

Also note that in this case only the first dash is useful, the closing one won't change anything.

Why can't we keep a simple rule here?

planctus:
Well, i think this doc would be helpful, gives you info on how to handle situations where you need to act to prevent whitespaces while keeping the syntax nice, it doesn't say you have to in any case.
About the complexity involved by this, i don't know, we are providing templates and their quality can be measured also by these specific choices we make, so, for how it's true there we might make mistakes, that it's in a way hacky to use a dash to mean "trim here", it's better than accepting the potential consequences of the lack of attention around the issue. 
We might then add other use cases to these examples, for sure, but i believe this is a good starting point.

yhuard:
I agree, one could forget to add the dash. But two? The QA process is here for that.

In your example, "over-trimming" is not an issue. But there are cases like the following one where over-trimming breaks the output:

```html
<input
  id="{{- _id -}}"
  name="{{- _name -}}"
  class="{{- _css_class -}}"
  {{- _disabled ? "disabled" -}}
  {{- _hidden ? "hidden" -}}
  {{- _extra_attributes|raw -}} 
>
```

If both  `_disabled` and `_hidden` are truthy, then the output will contain a `disabledhidden` attribute. 

Your solution (as discussed on slack) would be to add extra spaces like this, right?

```html
<input
  id="{{- _id -}}"
  name="{{- _name -}}"
  class="{{- _css_class -}}"
  {{- _disabled ? " disabled " -}}
  {{- _hidden ? " hidden " -}}
  {{- _extra_attributes|raw -}} 
>
```

Isn't it simpler to just no trim everything?

> Also note that in this case only the first dash is useful, the closing one won't change anything.

Line breaks are also white space characters. The closing whitespace control modifier will change something. 

```html
<div class="ecl-feedback-message ecl-feedback-message--something">
  {{- _invalid_text -}}
</div>
```

is equivalent to:

```html
<div class="ecl-feedback-message ecl-feedback-message--something"> {{- _invalid_text -}} </div>
```

emeryro:
we can give it a try with your set of rules, but I'm pretty sure that there will be lots of cases where we miss a dash (it's already quite complex to review to be honest)

yhuard:
Good to have in mind: https://www.drupal.org/docs/develop/coding-standards/twig-coding-standards

(still reading it)

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Decision
+
+In order to avoid issues related to whitespace, we should strip them by default and only add them when needed.
+
+## Context
+
+Whitespaces can be a source of problems in HTML when not handled correctly.
+
+When we develop our components in the [main repository](https://github.com/ec-europa/europa-component-library), we use React, which [removes most of the whitespaces](https://reactjs.org/docs/jsx-in-depth.html#string-literals-1.). We don't specifically think about whitespaces and how the can impact the end result, since there are none.
```
kalinchernev:
```suggestion
When we develop our components in the [main repository](https://github.com/ec-europa/europa-component-library), we use React, which [removes most of the whitespaces](https://reactjs.org/docs/jsx-in-depth.html#string-literals-1.). With such behavior from React, we don't think about whitespaces and how they can impact the end result after rendering.
```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+
+With Twig, we need to be careful about whitespaces. When they appear [between inline block elements](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) or around content, they alter the output.
```
kalinchernev:
```suggestion
However, with Twig, we need to be careful with whitespaces. When they appear [between inline block elements](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) or around content, they alter the output.
```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+</my-component>
+
+{% endspaceless %}
+```
+
+If you need to add a whitespace somewhere in your template, you can do it with: `{% endspaceless %}{{ ' ' }}{% spaceless %}`.
+
+#### Use the whitespace control modifier on your tags when needed
+
+You can trim leading and or trailing whitespaces with `-` (dash), e.g:
```
kalinchernev:
```suggestion
You can trim leading and trailing whitespaces with `-` (dash) in the following ways:
```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+
+- `{{- my_var }}`: remove left whitespace
+- `{{ my_var -}}`: remove right whitespace
+- `{{- my_var -}}`: remove both whitespaces
+
+It makes sense to use it when you print content between 2 tags:
+
+```html
```
kalinchernev:
```suggestion

Between adjacent tags:

```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+  {{- my_var -}}
+<span>
+
+<!-- Don't -->
+</span>
+  {{ my_var }}
+<span>
+```
+
+or inside a tag:
```
kalinchernev:
```suggestion
Or when inside a tag, i.e. between a tag and its children
```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+  {{- my_var -}}
+</span>
+
+<!-- Don't -->
+<span>
+  {{ my_var }}
+</span>
+```
+
+But it doesn’t make sense in attributes,
```
kalinchernev: An example why it doesn't make sense would be appreciated.

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+  ```html
+  <span></span> <span></span>
+  ```
+
+- between a tag and its children
+
+  Expect:
+
+  ```html
+  <span><span></span><span></span></span>
```
kalinchernev: Later in the document, a single `html` block contains `<!-- Do -->` and `, might worth making consistent.

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+<span>{{ my_var }}</span>
+
+<!-- Don't -->
+<span>{{- my_var -}}</span>
+```
+
+Notes:
+
+- `<span />` is used in these examples, but it could be any tag
+- `{{` can be replaced by any content-printing tag, e.g. `{% include '...' %}`, `{% embed '...' %}` or `{% block '...' %}`. In most cases, you don't need to remove the whitespaces on control structures (`if`, `for`).
```
kalinchernev:
```suggestion
- `{{` can be replaced by any content-printing tag, e.g. `{% include '...' %}`, `{% embed '...' %}` or `{% block '...' %}`. In most cases, you don't need to remove the whitespaces in control structures (`if`, `for`).
```

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Alternatives Considered
+
+### Single-line output
+
+We could try to get the output of the Twig template to be on a single line but it would mean stripping all the newlines in the elements attributes.
```
kalinchernev:
```suggestion
We could try to get the output of the Twig template to be on a single line but it would mean stripping all newlines from elements' attributes.
```

docs/decisions/001-strip-whitespaces.md
yhuard: instead of Expect/avoid, use Do/don't? is that it?

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,228 @@
...
+```html
+<select
+  {{- ' id="#{_id}"' -}}
+  {{- ' name="#{_name}"' -}}
+  {{- ' class="#{_css_class}"' -}}
+  {{- _disabled ? " disabled" -}}
+  {{- _extra_attributes|raw -}}
+>
+```
+
```
kalinchernev:
```suggestion

This approach is not feasible, because it makes the code harder to read and maintain, and it's also more error-prone.

```

docs/decisions/001-strip-whitespaces.md
yhuard: no sure about this specific change, wouldn't it break the markup here? (removing ```html)

docs/decisions/001-strip-whitespaces.md
kalinchernev: Both forms are ok, it would be easier to follow along the text if only one is used.

docs/decisions/001-strip-whitespaces.md
kalinchernev: I must have selected the wrong line, you're right.

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,235 @@
...
+##### Complex scenario: the content is surrounded by Twig tags
+
+Sometimes, you will find that control structures (`if`, `for`) and other Twig tags (like `set` or comments) may alter the output by introducing undesired whitespaces.
+
+As a rule of thumb, it's good to start by removing all the whitespaces for these tags using the dash syntax, generate the snapshot of the finished component and then start removing the dash where they're unnecessary.
+
+Here's the plan to follow:
+
+1. the developper creates the component X with dashes on all the control structures and other tags that might generate whitespaces
```
planctus: small typo, should be developer in the two lines ;)

docs/decisions/001-strip-whitespaces.md
```diff
@@ -0,0 +1,235 @@
...
+2. the developper takes the Jest snapshots and open a PR
+3. the reviewer validates the component and the snapshots, and then merges the PR
+4. the reviewer creates a Jira ticket: "Improve quality of component X". The goal of the ticket is to remove as many dashes as possible.
```
planctus: If i get it correctly the idea would be to apply blindly dashes to remove any potential whitespace, merge if the component looks god and after that creating a follow up ticket to verify if there's any chance to remove dashes here and there.
I haven't fully partecipated to the elaboration of this helpful document, but this flow looks inefficient to me, why don't we consider the usage of whitespace removal as any other "syntax" related thing? Ideally it should come as a "prettier" option, although it is something that might needs a case by case evaluation.

emeryro: the issue is that there is no simple rule to determine if we should put a dash.
Sometime you have to add it to a `if`, sometimes you don't

emeryro:
This document always seems very complex to me...
If we decide to go to the four points approach described, we may replace half of the document by: "put dashes everywhere in the print area of the twig template (control structure and variables), expect for attributes and variables with no space around"

docs/decisions/001-strip-whitespaces.md
yhuard: We can also write the whole code on a single line: this way, we are sure the code we produce is ok. React does it for us already. If we wanted to be 100% accurate on the ECL website, we should tell our beautifier to not add any newline and to not indent the code displayed in the showcase... but then, it wouldn't be really readable. 

Prettier describes it well I think: https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting 

The code produced by the twig template must work as expected. It should not introduce breaking whitespaces, as it was the case for the link component. The only way we have to ensure this is to remove all the whitespaces between an within HTML tags.

I hope this helps clarifying the context :)

This question of whitespaces is not only a matter of "syntax", it's not as choosing tabs vs spaces. 

`spaceless` is pretty good at removing whitespaces between HTML tags. The main issue we face is when we mix HTML tags and text. That's when things go bad...

degliwe:
My 2 cents.
- first rule: not everything should be a rule
- We should be crystal clear on the expected resulting markup with a couple more example (at least one with a real component).
- Then we expose the various strategies and tools to get to the expected result. (the page is already doing a good job at exposing them with their pro's and con's)
- Maybe some of the strategies can becomes rules (ex: use of spaceless)
- But we accept any component outputting proper markup (ex: build template wiht css in mind, one line template....)

docs/decisions/001-strip-whitespaces.md
```diff
@@ -11,7 +11,14 @@
 
 ## Decision
 
-In order to avoid issues related to whitespaces, we should strip whitespaces between HTML tags by default and only add them when needed. The Twig template is considered valid when the output it produces doesn't contain any whitespace between 2 HTML tags or between an HTML tag and its text content, unless the whitespace is explicitly required.
+In order to avoid issues related to whitespaces, we should strip whitespaces between HTML tags by default and only add them when needed.
+
+The Twig template is considered valid when the output it produces doesn't contain any whitespace:
+
+- between 2 HTML tags ;
```
kalinchernev: I personally feel these 2 rules/guidelines are already a good base for being able to move forward with the document and it's implementation.

PR #46, opened 2019-03-01
feat: Add expandable component

kalinchernev:
# PR description

I've tried to make the composition of the button structure easy to follow inside the template, by following a path of which variable is defined and later use how.

Although it's tempting to expose the `collapsed` property to the public, in order to allow for `knobs`, I haven't, following the story of the React component.

All buttons have been updated to include `data-ecl-label` attribute, probably something not [migrated](https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/react/packages/button/src/Button.jsx#L38) yet because of a lack of use case. Which happens in this pull request, as the vanilla JS implementation requires the presence of this attribute.

The `@ecl/ec-utility-typography` dependency is added here, but not in the React, because React can render children in a more flexible way, here we need it because of `ecl-u-type-paragraph-m`.

## QA Checklist

- [x] I have provided tests


src/ec/packages/ec-component-expandable/expandable.story.js
```diff
@@ -0,0 +1,49 @@
...
+
+storiesOf('Components/Expandable', module)
+  .addDecorator(withKnobs)
+  .addDecorator(withCode)
+  .addDecorator(withNotes)
+  .add(
+    'default',
+    () => {
+      const html = expandable({
+        label_expanded: text('Label Expanded', demoData.labelExpanded),
```
emeryro: as label is injected va js, the knobs do not work here.
I checked on ECL, we don't use knobs for expandable labels. So I guess we could just remove them.

src/ec/packages/ec-component-expandable/expandable.test.js
```diff
@@ -0,0 +1,18 @@
+import path from 'path';
+import { renderTwigFileAsNode } from '@ecl-twig/test-utils';
+
+import demoData from './demo/data';
+
+// Add SVG icon path.
+demoData.button.icon.path = 'example';
+
+describe('EC - Expandable', () => {
```
emeryro: it should be possible to add a snapshot for expanded state, to be sure that markup is correct every time.

src/ec/packages/ec-component-expandable/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "@ecl-twig/ec-component-expandable",
+  "author": "European Commission",
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Expandable",
+  "dependencies": {
+    "@ecl-twig/ec-component-button": "^0.0.1-alpha"
+  },
+  "devDependencies": {
+    "@ecl/ec-component-button": "^2.1.1",
```
emeryro: there is probably no need to have vanilla button as devDependency, as it is alreay the case in `ecl-twig/ec-component-button`

src/ec/packages/ec-component-expandable/expandable.story.js
kalinchernev: I managed to fix this by adding the storybook's methods in the `demoData` itself. If it's an issue to have knobs which were not there before in the React version, I can remove them.

src/ec/packages/ec-component-expandable/expandable.test.js
kalinchernev: I am two hands for having a test for expanded state! I tried the following so far:

1) I tried to `then` the `expect(render(demoData))` of the test (`await` yields declaration error) using similar approach from the way we include [DOM-related JS](https://github.com/ec-europa/ecl-twig/blob/master/src/ec/packages/ec-component-breadcrumb/breadcrumb.story.js#L57) in stories, and `var button = document.querySelector("[data-ecl-expandable-toggle=true]");` with `button.click` which would trigger the DOM state change, but `jest` does not do anything, of course, the script will go in the snapshot as a DOM contents, but won't run. For this to work, we need a headless browser, I guess.

2) I tried to approach the issue from using a `boolean` flag exposed to the public. Thus, aiming fro a "knob" and a way to immitate state by checks in twig template. It won't be pretty, and we'll need to discuss new properties. For instance, do we expose only 1 property `aria-expanded` which modifies the rest down the road, or do we expose properties separately?

We have the following elements to take into account in order to render expanded state without DOM manipulation, but by input data to the template:

- `aria-expanded` on button. This brings the question: is this flag enough as a public "API"
- button label inside `ecl-button__label` span
- `hidden` attribute on `ecl-expandable__content`. This brings the question: wouldn't it better to expose "extra_attributes" type of variable for content-related attributes?

kalinchernev:
@emeryro thanks for the review!
I updated the code to accept suggestions to the most part. I've left https://github.com/ec-europa/ecl-twig/pull/46#discussion_r262186168 as a question, I'm not sure how would be the best way to approach this.
I haven't changed the label as I don't know what to do with regards to spaces as well.

src/ec/packages/ec-component-expandable/expandable.test.js
emeryro: I think that your second option is valid (at least that's probably what I would have do).
If we expose a parameter `default_state` (or something like that), which will control the initial `aria-expanded` and `hidden` it should be good

kalinchernev: I made a few updates with the suggested approach. I'm afraid, however, the JS behaviors of changing labels is not in the snapshot.

src/ec/packages/ec-component-expandable/expandable.html.twig
```diff
@@ -0,0 +1,102 @@
+{#
+  Parameters:
+    - "id" (string) (default: '')
+    - "label_expanded" (string) (default: '')
+    - "label_collapsed" (string) (default: '')
+    - "label_collapsed" (string) (default: '')
+    - "expanded" (string) (default: '')
```
emeryro: there are other component where we handle boolean directly instead of strings (https://github.com/ec-europa/ecl-twig/blob/master/src/ec/packages/ec-component-language-list/language-list.html.twig for instance), and it seems to wroks correctly

src/ec/packages/ec-component-expandable/expandable.html.twig
```diff
@@ -0,0 +1,102 @@
...
+      {% set _content_hidden = 'hidden' %}
+    {% endif %}
+
+  {% endif %}  
+
+  {%- include '../ec-component-button/button.html.twig' with {
+    label: _label,
+    variant: 'secondary',
+    type: 'button',
+    icon: {
```
emeryro: is it mandatory to declare the empty icon here?

src/ec/packages/ec-component-expandable/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Expandable",
+  "dependencies": {
+    "@ecl-twig/ec-component-button": "^0.0.1-alpha"
+  },
+  "devDependencies": {
+    "@ecl/ec-component-expandable": "^2.1.1",
+    "@ecl/ec-specs-expandable": "^2.1.1",
+    "@ecl/ec-utility-typography": "^2.1.0"
```
emeryro: should be `^2.1.1` too

src/ec/packages/ec-component-expandable/expandable.story.js
```diff
@@ -0,0 +1,57 @@
...
+    'default',
+    () => {
+      demoData.button.icon.path = defaultSprite;
+      demoData.label_expanded = text('Label Expanded', demoData.labelExpanded);
+      demoData.label_collapsed = text(
+        'Label Collapsed',
+        demoData.labelCollapsed
+      );
+      demoData.content = text('Content', demoData.content);
+      demoData.expanded = radios(
```
emeryro: as the behavior is not fully functional (no label change, no changes of radio button if we click on the expandable component itself), I think that there is no need to display this knobs.
The boolean behind it is still useful in the twig template, but having it as a knob may raise questions.

src/ec/packages/ec-component-expandable/expandable.html.twig
kalinchernev: Good question! In fact, I don't even know why I've "enforced" an icon, it's an option item in the button. So I'll remove.

src/ec/packages/ec-component-expandable/package.json
kalinchernev: I must have copied that wrongly from somewhere, fixed.

src/ec/packages/ec-component-expandable/expandable.story.js
kalinchernev: Agree, changed.

src/ec/packages/ec-component-expandable/expandable.html.twig
kalinchernev: Found a way to implement booleans, I hope it's better.

src/ec/packages/ec-component-expandable/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Expandable",
+  "dependencies": {
+    "@ecl-twig/ec-component-button": "^0.0.1-alpha"
+  },
+  "devDependencies": {
+    "@ecl/ec-component-expandable": "^2.1.1",
+    "@ecl/ec-specs-expandable": "^2.1.1",
+    "@ecl/ec-utility-typography": "^2.1.1"
```
emeryro: we should have a devDependency to `@ecl/ec-resources-icons` too

emeryro:
I'm not sure I fully get it...
Indeed, the idea is to have a `block` for description, as you did.
But why did you remove the dashes from control structure? I guess that's what causes this extra line.

PR #48, opened 2019-03-04
feat: EC component accordion - INNO-1344

kalinchernev:
# PR description

## QA Checklist

- [x] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-accordion/accordion.html.twig
+++ b/src/ec/packages/ec-component-accordion/accordion.html.twig
@@ -8,7 +8,7 @@
           label: (string),
           icon: (predefined structure): see Icon component,
         },
-        content: (string),
+        content: (block),
       },
       ...
     ]
@@ -60,8 +60,9 @@
     ]
   } -%}
   </dt>
-  <dd class="ecl-accordion__content" id="{{- item.id -}}" hidden="" role="region">
-    {{- item.content -}}
+  <dd class="ecl-accordion__content" id="{{- item.id -}}" hidden role="region">
+    {% set _content = item.content|default('') %}
+    {%- block content _content -%}
   </dd>
   {%- endfor -%}
   {%- endif -%}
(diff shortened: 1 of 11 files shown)


src/ec/packages/ec-component-accordion/accordion.html.twig
```diff
@@ -0,0 +1,69 @@
...
+  Parameters:
+    - "items" (array) (default: []): format: [
+      {
+        "id": (string),
+        level: (string),
+        toggle: (predefined structure): see Button component, format: {
+          label: (string),
+          icon: (predefined structure): see Icon component,
+        },
+        content: (string),
```
emeryro: I don't know if that would be easy (or even possible), but that would be better to use blocks for accordion content, as it may not be simple text

src/ec/packages/ec-component-accordion/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "license": "EUPL-1.1",
+  "version": "0.0.1-alpha",
+  "description": "ECL Twig - EC Accordion",
+  "dependencies": {
+    "@ecl-twig/ec-component-button": "^0.0.1-alpha"
+  },
+  "devDependencies": {
+    "@ecl/ec-resources-icons": "~2.1.0",
+    "@ecl/ec-component-accordion": "^2.1.1",
+    "@ecl/ec-component-button": "^2.1.1",
```
emeryro: unneeded devDependency

src/ec/packages/ec-component-accordion/accordion.html.twig
kalinchernev: I made the change in a way to allow for a block in the place of the content if the item would contain one. I tried to think of an example, but couldn't reach a valid twig syntax.

I tried to override `item.content` in a loop with a block. I couldn't manage to change items of an iterable from one format to another. Please have a look at the latest change 540b81c and let me know if that was what you had in mind.

If you have a scenario how we can test that in a composition, please let share.

PR #50, opened 2019-03-05
feat: Media container 

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-media-container/media-container.test.js
```diff
@@ -0,0 +1,62 @@
+import path from 'path';
+import { renderTwigFile } from '@ecl-twig/test-utils';
+
+describe('EC - Media container image', () => {
+  const template = path.resolve(__dirname, './media-container.html.twig');
+  test(`- Media container image renders correctly`, done => {
+    expect.assertions(1);
+
+    const defaultDataStructure = {
```
kalinchernev: Please use adapter spec data. If spec data changes, tests will miss it otherwise.

src/ec/packages/ec-component-media-container/media-container.test.js
```diff
@@ -0,0 +1,62 @@
...
+    });
+  });
+});
+
+describe('EC - Media container video', () => {
+  const template = path.resolve(__dirname, './media-container.html.twig');
+  test(`- Media container video renders correctly`, done => {
+    expect.assertions(1);
+
+    const defaultDataStructure = {
```
kalinchernev: Same here.

src/ec/packages/ec-component-media-container/media-container.test.js
```diff
@@ -0,0 +1,62 @@
...
+    renderTwigFile(template, defaultDataStructure, (err, html) => {
+      expect(html).toMatchSnapshot();
+      done();
+    });
+  });
+});
+
+describe('EC - Media container video', () => {
+  const template = path.resolve(__dirname, './media-container.html.twig');
+  test(`- Media container video renders correctly`, done => {
```
kalinchernev: Although I'm a fan of back ticks, I don't think they are actually needed.

src/ec/packages/ec-component-media-container/media-container.html.twig
```diff
@@ -0,0 +1,75 @@
...
+{% if extra_attributes is defined and extra_attributes is not empty and extra_attributes is iterable %}
+  {% for attr in extra_attributes %}
+    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value ~ '"' %}
+  {% endfor %}
+{% endif %} 
+
+{# Print the result #}
+
+<figure class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+{%- if _sources is not empty -%}
```
kalinchernev: Might worth transferring the checks of `iterable` too (https://github.com/ec-europa/europa-component-library/blob/v2-dev/src/systems/ec/implementations/react/packages/media-container/src/MediaContainer.jsx#L18)

src/ec/packages/ec-component-media-container/media-container.html.twig
```diff
@@ -0,0 +1,75 @@
...
+  <video class="ecl-media-container__media" controls="controls" poster="{{ _image }}" alt="{{ _alt }}">
+  {%- for source in _sources -%}
+    <source type="{{ source.type }}" src="{{ source.src }}" />
+  {%- endfor -%}
```
kalinchernev: Please consider if would actually need to remove white spaces in cases such as this.

src/ec/packages/ec-component-media-container/media-container.html.twig
```diff
@@ -0,0 +1,75 @@
...
+{%- if tracks is not empty -%}
+  {%- for track in _tracks -%}
+    <track kind="{{ track.kind }}" srcLang="{{ track.srcLang }}" label="{{ track.label }}" src="{{ track.src }}" />
```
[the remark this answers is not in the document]
planctus: mmhh...i see, but in this case this property (srcLang) is coming from ECL, is not anything "defined" here, not sure it would be worth altering the parameters of the component in twig because of this.
What do you think?

planctus:
@kalinchernev, thanks for the review, i should have (hopefully) covered all the points but the one regarding the "snake case", i put a comment about it...

src/ec/packages/ec-component-media-container/media-container.html.twig
planctus: Ah, i have to say that i'm confused, after all, regarding how to handle this, at the moment i removed probably all the dashes from control structures and it doesn't look like we would need to trim anything else but i might be totally wrong.

src/ec/packages/ec-component-media-container/media-container.html.twig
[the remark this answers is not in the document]
kalinchernev: With my experience so far, all variables which do not follow snake case get reformatted before the twig template. And twig templates accept only snake case. And we need to adapt that before data is passed to the twig template, like this [example](https://github.com/ec-europa/ecl-twig/blob/master/src/ec/packages/ec-component-social-media-follow/demo/data.js#L11).

src/ec/packages/ec-component-media-container/media-container.html.twig
```diff
@@ -0,0 +1,75 @@
...
+{% if _sources is not empty and sources is iterable %}
+  <video class="ecl-media-container__media" controls poster="{{ _image }}" alt="{{ _alt }}">
+  {% for source in _sources %}
+    <source type="{{ source.type }}" src="{{ source.src }}" />
+  {% endfor %}
+{% if tracks is not empty and tracks is iterable %}
+  {% for track in _tracks %}
+    <track kind="{{ track.kind }}" srclang="{{ track.srcLang }}" label="{{ track.label }}" src="{{ track.src }}" />
+  {% endfor %}
+{% endif %}
```
kalinchernev: It's tricky closing the if statement and the video tag after it at this place. I would suggest changes in indentation:

```twig
<figure class="{{ _css_class }}"{{ _extra_attributes|raw }}>
  {% if _sources is not empty and sources is iterable %}
    <video
      alt="{{ _alt }}"
      class="ecl-media-container__media"
      controls
      poster="{{ _image }}"
    >
    {% for source in _sources %}
      <source type="{{ source.type }}" src="{{ source.src }}" />
    {% endfor %}
    {% if tracks is not empty and tracks is iterable %}
      {% for track in _tracks %}
        <track
          kind="{{ track.kind }}"
          label="{{ track.label }}"
          src="{{ track.src }}"
          srclang="{{ track.srcLang }}"
        />
      {% endfor %}
    {% endif %}
    </video>
  {% else %}
    <img
      alt="{{ _alt }}"
      class="ecl-media-container__media"
      src="{{ _image }}"
    />
  {% endif %}
  <figcaption class="ecl-media-container__caption">
    {{ _description }}
  </figcaption>
</figure>

I've placed attributes beneath each other to keep the code to under 80 chars ruler if we had one.

PS, not required to accept this, but please indent if/endif statements to be easy to see on first look, thanks.

src/ec/packages/ec-component-media-container/media-container.html.twig
kalinchernev: I'd personally do the same. Except, please trim `{{ _description }}` as it's definitely a place where whitespaces won't be good.

PR #56, opened 2019-03-07
feat(tag): improve template quality

yhuard:
This PR removes the unnecessary dashes and apply the missing `only`.

Jira: INNO-1369


emeryro:
Modifications are good, but now that we removed the "button tag" from ECL, I guess we should remove it from there too

PR #58, opened 2019-03-07
refactor(icon): improve quality, add more tests

yhuard:
Unnecessary condition removed, more tests added.

Jira: INNO-1362


yhuard:
@kalinchernev I've refactored the code after your first remark.

Regarding the second one, I think it will add more complexity to the code, unless you have a simple solution in mind?

PR #60, opened 2019-03-07
refactor(accordion): improve quality of the template

kalinchernev:
Jira: INNO-1356


src/ec/packages/ec-component-accordion/accordion.html.twig
```diff
@@ -44,29 +44,38 @@
...
+          ]
+        } only %}
+      </dt>
+      <dd
+        class="ecl-accordion__content"
+        hidden
+        id="{{ item.id }}"
+        role="region"
+      >
+        {%- set _content = item.content|default('') -%}
```
yhuard: almost perfect, but I think there's one more dash you can remove, either the ending one `-%}` from the `set` tag, or the starting `{%-` one from the `block`. In general, I remove the ending one

PR #62, opened 2019-03-07
chore: update to storybook 5

yhuard:


emeryro:
The new look and feel is indeed much better!
New options will makes life easier (responsive display, zoom, fullscreen, ...)
I just found a little bug with keyboard navigation (going to previous story doesn't seems to work). Keyboard navigation doesn't work on IE by the way.

The search box is also actiong strangely... If I type `se` i get only one result, if I type `sea` I get lots of results, and not really related.

But anyway, no blocker here.

yhuard:
Indeed, keyboard navigation is far from great. The search is not ideal either. Let's hope it will be fixed soon :) or make PRs on storybook 😉

PR #63, opened 2019-03-08
feat: render twig templates with PHP - INNO-1375

kalinchernev:
# PR description

I've tried to make the setup accessible by adding root-level script linking to package's script, linking to composer's script. Nothing new to learn for the user.

There is a custom Twig loader facilitating path resolution. In overall, it's based on files within `node_modules`. Scripts are system-oriented, set by `ECL_SYSTEM`.

The logical workflow of the renderer is like following:
- Assumes data for templates is stored with `demo` folder in npm packages. It seemed the closes assumption I could make based on existing patterns within ECL's main repository and this one.
- Tries to find a local override of data with in the package stored in this repository.
- If the previous is not present, it tries to fallback to ECL's main spec.
- Several data files per component are supported, i.e. when variants. (check Button)
- If neither of the two previous attempts do not succeed, pass empty data and parse anyway. (check Breadcrumb)


.gitignore
```diff
@@ -1,4 +1,7 @@
+.idea
```
yhuard: `.idea` is specific to your environment, isn't it? it should be in your global .gitignore, no in the one of the project :) https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

kalinchernev: I don't mind treating this as a global to my environment, though I doubt others will do the same.
It's created by PHPStorm.

yhuard: Others should do the same. Isn't it one of the Git best practices? The project `.gitignore` should only ignore files and folders specific to the project. `.idea`, `.DS_Store`, `*.sublime-workspace`, all of those are specific to the user's environment and should be ignored in the global, user-specific `.gitignore`, shouldn't they?

src/php/packages/renderer/README.md
```diff
@@ -0,0 +1,72 @@
...
+```shell
+$ yarn setup
+```
+
+```shell
+$ composer run setup
+```
+
+Note that the `npm` scripts facilitate integration between tools in Node.js and PHP. Scripts naming may differ because of reserved keywords, such as `install` in composer, which is replaced by `setup`.
+
```
emeryro: We should also have a paragraph here to explain how to clean the data from previous build (as in getting-started.md)

PR #64, opened 2019-03-08
refactor(blockquote): update dependency

yhuard:
Not much to improve here.

Jira: INNO-1357


kalinchernev:
As discussed in slack, organization of documentation is ok here. Other components which have several variants would have `/docs`.

PR #68, opened 2019-03-08
refactor(language-list): improve quality

yhuard:
Jira: INNO-1363

### DIFF
--- a/src/ec/packages/ec-component-language-list/language-list-item.html.twig
+++ b/src/ec/packages/ec-component-language-list/language-list-item.html.twig
@@ -19,10 +19,10 @@
 
 {# Internal logic - Process properties #}
 
-{% set _item_icon = 
+{% set _item_icon =
   _item.active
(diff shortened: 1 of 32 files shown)


src/ec/packages/ec-component-language-list/language-list-item.html.twig
```diff
@@ -19,10 +19,10 @@
 
 {# Internal logic - Process properties #}
 
-{% set _item_icon = 
+{% set _item_icon =
```
kalinchernev: Could you please double-check also the spacing between brackets?
- https://github.com/ec-europa/ecl-twig/pull/68/files#diff-4f104de808bce416807a253fbf41b36dR35
- https://github.com/ec-europa/ecl-twig/pull/68/files#diff-4f104de808bce416807a253fbf41b36dL36
- https://github.com/ec-europa/ecl-twig/pull/68/files#diff-4f104de808bce416807a253fbf41b36dR38

PR #74, opened 2019-03-11
feat: EC pagination - INNO-1355

emeryro:
# PR description

Add component pagination (EC only)

## QA Checklist

- [x] I have provided tests


src/ec/packages/ec-component-pagination/pagination.html.twig
```diff
@@ -0,0 +1,78 @@
...
+    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value ~ '"' %}
+  {% endfor %}
+{% endif %}
+
+{# Print the result #}
+
+<nav class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  <ul class="ecl-pagination__list">
+    {% for _item in _items %}
+      <li class="ecl-pagination__item{{ _item.current ? ' ecl-pagination__item--current' }}{{ _item.previous ? ' ecl-pagination__item--previous' }}{{ _item.next ? ' ecl-pagination__item--next' }}">
```
planctus: Couldn't we have a smarter way of handling this? I just mean that we have a correspondence between the keys in the "link item" (current, previous, next) and the name of the modifier classes, maybe we could set the base class and then add the modifier when the key is not empty instead of having to explicitely set them here.
Just an idea..

PR #75, opened 2019-03-11
feat: EC page banner - INNO-1379

emeryro:
# PR description

Add component page banner (EC only)

## QA Checklist

- [x] I have provided tests


src/ec/packages/ec-component-page-banner/page-banner.html.twig
```diff
@@ -0,0 +1,67 @@
...
+    <div class="ecl-page-banner__image" style="background-image:url({{ _image }})"></div>
+  {% endif %}
+  <div class="ecl-container ecl-page-banner__container">
+    <div class="ecl-page-banner__content">
+      <h1 class="ecl-page-banner__title">{{ _title }}</h1>
+      {% if _baseline is not empty %}
+        <p class="ecl-page-banner__baseline">{{ _baseline }}</p>
+      {% endif %}
+      {% if _button is not empty %}
+        {%- include '../ec-component-button/button.html.twig' with _button|merge({
```
yhuard: dashes are not needed here

src/ec/packages/ec-component-page-banner/.npmignore
```diff
@@ -0,0 +1,5 @@
+__snapshots__
+docs
```
yhuard: line not needed

src/ec/packages/ec-component-page-banner/.npmignore
```diff
@@ -0,0 +1,5 @@
...
+*.story.js
+*.test.js
+**/*.md
```
yhuard: actually we want the main README.md file to be published on npmjs.com, thus we don't want to ignore it ;)

src/ec/packages/ec-component-page-banner/page-banner.html.twig
emeryro: I guess I will never understand when we have to put dashes or not...

src/ec/packages/ec-component-page-banner/.npmignore
emeryro: ok, but we will have to update `.npmignore` in other compoonents, as it is the same

src/ec/packages/ec-component-page-banner/page-banner.html.twig
yhuard: here you don't need them because the `include` will print a new HTML tag, and is surrounded by either `</p>` or `</h1>` before, and `</div>` after -> `spaceless` will remove the whitespaces between the HTML tags -> 🎉

PR #77, opened 2019-03-19
refactor(link and social): link supports array of icons

planctus:
# PR description

## QA Checklist

- [x] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-footer/__snapshots__/footer.test.js.snap
+++ b/src/ec/packages/ec-component-footer/__snapshots__/footer.test.js.snap
@@ -13,9 +13,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
     >
       Go to top
     </span>
-    
-    
-   
+     
     <svg
       aria-hidden="true"
       class="ecl-icon ecl-icon--fluid ecl-link__icon"
@@ -103,8 +101,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
                     xlink:href="static/icons.svg#branded--facebook"
                   />
                 </svg>
-                   
-      
+                 
                 <span
                   class="ecl-link__label"
                 >
@@ -128,8 +125,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
                     xlink:href="static/icons.svg#branded--twitter"
                   />
                 </svg>
-                   
-      
+                 
                 <span
                   class="ecl-link__label"
                 >
@@ -149,9 +145,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
                 >
                   Other social networks
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
@@ -188,9 +182,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
                 >
                   European Union
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
@@ -214,9 +206,7 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
                 >
                   EU institutions
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
@@ -329,9 +319,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
     >
       Go to top
     </span>
-    
-    
-   
+     
     <svg
       aria-hidden="true"
       class="ecl-icon ecl-icon--fluid ecl-link__icon"
@@ -419,8 +407,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
                     xlink:href="static/icons.svg#branded--facebook"
                   />
                 </svg>
-                   
-      
+                 
                 <span
                   class="ecl-link__label"
                 >
@@ -444,8 +431,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
                     xlink:href="static/icons.svg#branded--twitter"
                   />
                 </svg>
-                   
-      
+                 
                 <span
                   class="ecl-link__label"
                 >
@@ -465,9 +451,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
                 >
                   Other social networks
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
@@ -504,9 +488,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
                 >
                   European Union
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
@@ -530,9 +512,7 @@ exports[`EC - Footer Corporate renders correctly with extra attributes 1`] = `
                 >
                   EU institutions
                 </span>
-                
-    
-   
+                 
                 <svg
                   aria-hidden="true"
                   class="ecl-icon ecl-icon--xs ecl-link__icon"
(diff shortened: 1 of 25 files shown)


src/ec/packages/ec-component-footer/__snapshots__/footer.test.js.snap
```diff
@@ -48,20 +48,28 @@ exports[`EC - Footer Corporate renders correctly 1`] = `
               class="ecl-footer__section-item"
             >
               <a
-                class="ecl-link ecl-link--standalone ecl-footer__section-link"
+                class="ecl-link ecl-link--standalone ecl-link--icon ecl-link--icon- ecl-footer__section-link"
```
emeryro: When there is no icon, link should remain with a simple structure (no nested span)

src/ec/packages/ec-component-link/link.html.twig
```diff
@@ -40,16 +40,19 @@
   size: ''
 }%}
 
-{% if icon is defined %}
+{% if icon.name is defined and icon.name is not empty %}
   {% set _icon = _icon|merge(icon) %}
+{% elseif icon is defined and icon is not empty and icon.name is not defined %}
```
emeryro: Just a thought: isn't it easier to check first of there is something in `icon`, then check if it is an array?

src/ec/packages/ec-component-link/link.html.twig
```diff
@@ -68,25 +71,49 @@
...
-        extra_classes: 'ecl-link__icon'
-      } only %}
-      {{- '&nbsp;' }}
-    {%- endif -%}
+  {%- if _icons is defined and _link.icon_position == 'before' -%}
+    {%- for icon in _icons -%}
+      {%- if icon.extra_classes is defined and icon.extra_classes is not empty -%}
+        {%- set icon_extra_classes = icon.extra_classes ~ ' ' ~ icon_extra_classes -%}       
+      {%- endif -%}
+      {%- set _icon = _icon|merge(icon) -%}
```
emeryro: Each icon should already have the correct structure, isn't it? So why not using them directly in the include instead of using a temporary variable?
Also, I'm not really sure about it, but won't the second icon inherit all the properties of the first one, as the variable is not reset?

src/ec/packages/ec-component-link/link.html.twig
```diff
@@ -68,25 +71,49 @@
...
-        extra_classes: 'ecl-link__icon'
-      } only %}
-    {% endif %}
-  {% else %}
-    {{- _link.label }}
-  {%- endif %}
+  {%- endif -%}
+  {%- if _icons is defined and _link.icon_position == 'after' -%}
+    {{- '&nbsp;' -}}
+    {%- for _icon in _icons -%}
```
emeryro: It may be cleaner to put the markup of the icons on a variable once instead of duplicating the code

src/ec/packages/ec-component-social-media-follow/__snapshots__/social-media-follow.test.js.snap
```diff
@@ -144,23 +150,177 @@ exports[`EC - Social Media Follow Horizontal renders correctly 1`] = `
...
+          <use
+            xlink:href="example#gplus_hover"
+          />
+        </svg>
+         
         <span
           class="ecl-link__label"
         >
-          Other social networks
+          Google+
```
emeryro: As we are there, we can already replaced Google+ network by Instagram (we just did it on ECL)

src/ec/packages/ec-component-link/link.html.twig
planctus: The fact that there's no direct way of distinguishing an object from an array of objects (iterable works with both) is the reason of this weird logic here, i have to say that i've been trying diffferent approaches but probably not the easiest one, which would be checking for `icon[1] is not empty` to identify the array use case.

src/ec/packages/ec-component-footer/__snapshots__/footer.test.js.snap
planctus: Mmhh..yes but then why do we call the link component passing an icon if we don't need it?
The way the icon is passed here (only the size) tricks the logic in the link component. 
The thing here is: what should be the miminum requirement for outputting an icon in the link component? I think we had a check for the icon name before, i can set it back, but still is questionable the fact that we pass data we don't need, i think.

src/ec/packages/ec-component-link/link.html.twig
planctus: Yeah, i actually think that we might totally remove the distinction between the single and the multiple icon in the logic, what i mean is that i could probably just put the single one in an array and then we would just have two for cycles (before and after) with one include of the link component each.

src/ec/packages/ec-component-social-media-follow/__snapshots__/social-media-follow.test.js.snap
planctus: oooook

src/ec/packages/ec-component-link/link.html.twig
planctus: I also see that i haven't done this consistently.
The extra_classes have to be extracted, but for the rest it looks like i lost a bit the focus with all this icon _icon _icons :)

src/ec/packages/ec-component-footer/__snapshots__/footer.test.js.snap
emeryro: Indeed we did a test on the icon name. I don't have better idea for this, so I guess we could just stick to this test.
The issue with the footer, it that icon size are enforced, to looks like ECL's footer. We could indeed set the icon size in the data file, but in this case that would be up to the integrator to set the correct size (with potential errors)

src/ec/packages/ec-component-link/link.html.twig
emeryro: indeed, `is iterable` isn't smart enough in this case. There seems to be different ways, like [this](https://stackoverflow.com/questions/13876023/check-if-variable-is-string-or-array-in-twig/29642662#29642662), but that's not really simpler (and I'm not sure we could do it in our case)

src/ec/packages/ec-component-social-media-follow/__snapshots__/social-media-follow.test.js.snap
emeryro: we will release ECL 2.2.0 this afternooon, so we can already do it here

src/ec/packages/ec-component-link/__snapshots__/link.test.js.snap
```diff
@@ -31,7 +40,7 @@ exports[`EC - Link With icon after renders correctly 1`] = `
    
   <svg
     aria-hidden="true"
-    class="ecl-icon ecl-icon--fluid ecl-link__icon"
+    class="ecl-icon ecl-icon--fluid ecl-link__icon ecl-test-extra-class"
```
emeryro: there shouldn't be an extra class here I guess, as it is not in the test description

src/ec/packages/ec-component-link/link.test.js
```diff
@@ -43,6 +43,134 @@ describe('EC - Link', () => {
...
   });
 
+  describe('Standalone with missing icon name', () => {
+    const options = merge(defaultDataStructure, {
+      link: {
+        type: 'standalone',
+        label: 'Standalone link with icon',
+        icon_position: 'after',
+      },
+      icon: [
```
emeryro: If I'm not wrong, you could just pass the icon object instead of wrapping it in an array

src/ec/packages/ec-component-link/__snapshots__/link.test.js.snap
planctus: why not? it looks like it is defined at line 209 of link.test.js
`icon: {
        type: 'ui',
        name: 'external',
        size: 'fluid',
        path: defaultIconPath,
        extra_classes: 'ecl-test-extra-class',
      }
`

src/ec/packages/ec-component-link/link.test.js
planctus: Yep, sure ;)

src/ec/packages/ec-component-link/__snapshots__/link.test.js.snap
emeryro: Ok, I thought that it was a parameter leak

src/ec/packages/ec-component-social-media-follow/__snapshots__/social-media-follow.test.js.snap
planctus: How? :)

emeryro:
It seems to be good now.
Could you just fix the conflict and make jest happy?

PR #84, opened 2019-03-22
feat: render twig templates with PHP - INNO-1375

kalinchernev:
Builds upon #63.

Improvements:
- Uses `@babel/node` in order to resolve commonjs and ES6 modules. No need to convert current data for stories into commonjs.
- More graceful on lack of data. Does not stop rendering components after missing.
- Data generation script is project-wide, not PHP-specific anymore.
- Data and HTML output are moved to `dist/` folder, already ignored from Git.
- Both PHP and JavaScript outputs are stored for comparison.


planctus:
Hi @kalinchernev, i spent a bit of time on this, first to understand the approach and then trying to see whether we could make any "real" use of these files we generate. 
I've found that mostly due to the recent addition of components as the result of multiple twig templates, there were warning coming out of the rendering script.
So i took the chance to understand what these warnings were about:
- components defined via multiple twig templates
- components with "issues" in code or in the demo files.

I didn't want to pollute this work, so i pushed the changes i made to fix these issues here:
https://github.com/ec-europa/ecl-twig/pull/87
In this branch i've also created a "storybook" instance to "demo" the components as rendered by php, this has many limitations at the moment and it doesn't play too well with the "multiplications" of the components (we demo each variant individually..).
But at the same time, some of the result of this "game" could be really useful to identify problematic or critical situations, we already saw that rendering through php was indeed exposing issues that we could not reproduce via twig.js

Not sure we really want to merge this implementation for now in master, but we could merge these "two" branches about php and use it to investigate some of the "failures", like components rendered without any content from the demo files, which might reveal something about the components.

PR #100, opened 2019-06-20
feat(accordion): add accordion 2 - INNO-1548

emeryro:
# PR description

Add second version of accordion (ECL >= 2.6.0)


src/ec/packages/ec-component-accordion2/accordion2.html.twig
```diff
@@ -0,0 +1,92 @@
...
+  {% for attr in extra_attributes %}
+    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value ~ '"' %}
+  {% endfor %}
+{% endif %}
+
+{# Print the result #}
+
+<div
+  class="{{ _css_class }}"{{ _extra_attributes|raw }}
+  data-ecl-accordion2="true"
```
yhuard: actually, you don't need `="true"`, only `data-ecl-accordion2`

src/ec/packages/ec-component-accordion2/accordion2.html.twig
```diff
@@ -0,0 +1,92 @@
...
+            <span class="ecl-accordion2__toggle-flex">
+              {% include '../ec-component-icon/icon.html.twig' with {
+                icon: {
+                  path: _item.toggle.icon.path,
+                  type: "ui",
+                  name: "plus",
+                  size: "m"
+                },
+                extra_classes: 'ecl-accordion2__toggle-icon',
+                extra_attributes: [{ name: "data-ecl-accordion2-icon", value: "true" }]
```
yhuard: it would be great if we didn't have to pass a `value` to extra attributes (unnecessary), but this change of logic would need to be applied to every components... unless we create some global macros

what do you think?

emeryro: actually, it does seems to work by putting only a name (without changing the logic). I'll update it

PR #110, opened 2019-07-19
chore: use ECL 2.9.0 - INNO-1661

emeryro:
# PR description

Update CSS/js and main version number

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "2.9.0",
+  "version": "2.9-dev",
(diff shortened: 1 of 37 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "2.8.0",
+  "version": "2.9.0",
```
yhuard:
```suggestion
  "version": "2.9-dev",
```

emeryro:
I can update the dependencies already, indeed...

Issue #126, opened 2019-09-11
[Security] Disable relative paths loading


drupol:
Hi all,

Some comments are relying on relative paths to load component into a component.

Ex: https://github.com/ec-europa/ecl-twig/blob/master/src/ec/packages/ec-component-button/button.html.twig#L58

I just had a discussion with the core Twig developers and apparently loading relative paths from a Twig template is not supported.
This is why Twig asks to define the paths where to look for templates.

This is apparently also for security matters.

Therefore, I would ask you guys if it would be possible to change from

```
      {% include '../ec-component-icon/icon.html.twig' with {
```

to

```
      {% include 'ec-component-icon/icon.html.twig' with {
```

Once this is done, we will be able to provide a Symfony bundle including the different flavors of ECL using Twig.

Thx

ademarco:
We would also prefer not to have relative path in Twig templates. Anyway, until this is fixed, the [ECL Twig Loader](https://github.com/openeuropa/ecl-twig-loader) project can be used, it's `dev-master` supports ECL Twig way of using relative paths.

drupol:
Thanks, we already evaluated it but here the idea is to not rely on another tool and to fix issue at the root directly.

[ECL Twig Loader](https://github.com/openeuropa/ecl-twig-loader) would have been an option but it requires to change the default Twig loader, or replace it with a Chain loader and add it there. To be honest, this is too much of a hassle already.

I'd like to hear from the maintainers what they think about this.

Thx

PR #128, opened 2019-09-12
fix: Issue #126: Disable relative paths by using Twing instead of TwigJS.

drupol:
# PR description

See #126 and https://github.com/twigjs/twig.js/issues/657 for some contexts.

TLDR;

This project uses a Twig implementation that is different (_and then not compliant_) with the official and original implementation from Symfony.
This prevent the adoption by other framework (_or fiddling with third party loader_) because specific function are not behaving like the original ones and prevent the use and might be introducing security concerns.

This PR:

- [x] Replace [Twig](https://www.npmjs.com/package/twig) with [Twing](https://www.npmjs.com/package/twing) to prevent using relative paths in templates and be totally compliant with the original Twig.
- [x] Twing carefully follows the original [Twig](https://twig.symfony.com/) implementation.
- [x] This will foster the adoption of this project by **a greater** number of applications.
- [x] This prevent security issues that relative paths loading might bring.

## QA Checklist

- [ ] I have provided tests


planctus:
@drupol, thanks a lot for the pull request and the incentive to improve the library in general.
I have to say that, generally speaking, there is no "official" or "original" implementation of twig in js, quoting the landing page of twing:

> Twing aims at implementing Twig specifications perfectly, without compromise. This is not an easy task due to the nature of Twig specifications: they don't exist officially and can only be deduced from the public documentation, the source code documentation and the test suite of the PHP reference implementation. It sometimes happens that something that was not part of either the documentations or the test suite suddenly becomes part of the specifications like the filter tag or the macros rework issues, putting Twing and all other non-reference implementations in the uncomfortable position of having to deal with a potential breaking change. Since Twig's team doesn't plan on releasing some official specifications for the language, we can't expect the problem to be solved anytime soon.

At the same time we knew that the solution we had for the template inclusion was "sub-optimal", although usable by the first consumers of this library.
We were also amazed by how easily you could replace twig and the twig-loader making it work almost flawlessly.
Despite this, i have some concerns:
- I've tried the latest versions of twing and twing-loader (3.x) and i had issues with the rendering of the components, while with the versions you used everything works fine. This didn't make me that confident.
- I would have expected the new approach to be different, meaning that i thought the loader would work with the template name only knowing where to find it, but here we are still forced to include the component with its containing folder name, basically concatenating the environemtal path we set with the one in the include.
- I wanted to adapt the tests to use the new "environment" so to better understand if we had any difference in the rendering by the two libraries.
You can run the tests yourselves, you'll see most of them failing due to the "extra_attributes" and html attributes in general, the twing rendering is duplicating double quotes everywhere or adding them where they are not expected to be, the snapshots you check the rendering against were taken with twig.js.
Regarding this, anyway, we were already reasoning around html attributes and i think this exercise was useful to better identify inconsistencies that we have, i created an issue https://github.com/ec-europa/ecl-twig/issues/165 to come out with a refined solution for this.

For the time being, then, i don't think we can move to twing, because of the attributes but also because if we really want to improve the way we load templates i think we should be more ambitious, we talked about namespaces, for instance, even though i don't necessarily see that approach perfectly matching our needs either.
Thanks again ;)

drupol:
Oh sad news. I will give it a try this weekend during the [EUFOSSA Hackathon](https://eufossahackathon.bemyapp.com/) and see what I can do.

Also, perharps that @ericmorand (_the Twing maintainer_) has something to say about this ?
It could be nice to have its point of view.

ericmorand:
@planctus:

> I would have expected the new approach to be different, meaning that i thought the loader would work with the template name only knowing where to find it, but here we are still forced to include the component with its containing folder name, basically concatenating the environemtal path we set with the one in the include.

The filesystem loader takes a parameter at construct that is the "root" path where all the include will be resolved from. I think it is what you are looking for but I may be wrong. By the way, this is something that is coming from TwigPHP: their filesystem loader works exactly this way. Twing also comes with a relative filesystem loader that resolve includes from the path of the template where the include is called. But what you need here is the non-relative filesystem loader - aka TwingLoaderFilesystem. For example `new TwingLoaderFilesystem('/my/environmental/path')` would resolve all the includes from `/my/environmental/path` so `{{ include('foo.twig') }}` would load the template `/my/environmental/path/foo.twig`.

> For the time being, then, i don't think we can move to twing, because of the attributes but also because if we really want to improve the way we load templates i think we should be more ambitious, we talked about namespaces, for instance, even though i don't necessarily see that approach perfectly matching our needs either..

`TwingFilesystemLoader` supports namespaces fully, just like to TwigPHP filesystem loader, using the same logic and same syntax. And writing custom loaders is very easy thanks to an explicit API coming from a TypeScript interface. I can help there, with great pleasure.

> You can run the tests yourselves, you'll see most of them failing due to the "extra_attributes" and html attributes in general, the twing rendering is duplicating double quotes everywhere or adding them where they are not expected to be, the snapshots you check the rendering against were taken with twig.js.

I'll take a look at that. Twing doesn't render anything that is not part of the passed context - so it should not duplicate double quotes. Not sure what happens here but I'll check to see if it's a bug in Twing.

planctus:
hey, thanks @ericmorand for all the explanations, for the loader it totally depends on how our library is organised, indeed, we have twig templates nested in each component's folder and at the moment i believe we would need to pass all these folders as an array in a `TwingFilesystemLoader` instance to get what we expect, we consider this possibility together with other alternatives like using a "namespaced" approach.
The issue we had with attributes was really consistent, meaning that if we say this:
```
{% for attr in extra_attributes %}
    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value ~ '"' %}
{% endfor %}
```
and the template receives the extra_attributes as an input:
```
        extra_attributes: [
          { name: 'data-test', value: 'data-test-value' },
          { name: 'data-test-1', value: 'data-test-value-1' },
        ],
```
This extra attributes will always have duplicated double quotes in our screenshots made with [jest](https://jestjs.io/docs/en)

`+   data-test=""data-test-value""`

 We'll keep an eye on twing and we keep a branch active with it, anyway, thanks again ;)

drupol:
Thanks for the explanation @planctus , let's hope we find a solution all together !

planctus:
Hi @ericmorand, hi @drupol, so...
first of all thanks, this turned out to be a sort of laundry for our project, many things have been investigated and revealed at some point some weakness in our code, so that many fixes have been possible thanks to this "experiment".

Last time i wrote you i reported something that was not entirely true, indeed @ericmorand that part of the code we use to prepare the extra attributes was not the problematic one, it was instead when we were rendering them not preventing the auto-escape, so simply like 
`{{ _extra_attributes }} `
applying the raw filter moved many of these duplicated quotes away, we had an inconsistent usage of the raw filter for the attributes so this was not always producing the duplicates.
The other use case was when we concatenated in a variable the value of another variable in a string.
`{{ _id ? 'id="' ~ _id ~ '"' }}`
and for how compact this syntax is i don't really like it, writing it instead like this:
```
{% if _id %}
    id="{{ _id }}"
 {% endif %}
```
works fine.

So, that's for the attributes, for the loader i did what i had in mind, just passing all the folders in which we have templates (this will have to be improved, but ok..) and now we can simply load components like:
`{% include 'icon.html.twig' with {} %}`
which is what we always wanted to do.

Other things were found along the way, one of this was the problematic usage of the default filter for a boolean setting a true value.

https://github.com/ec-europa/ecl-twig/pull/164/files
digging into the tests you would notice that we actually had bugs here and there that we (hopefully) fixed by doing this work.
Now i'm writing on a friday afternoon, things look very promising at this point, i still have a very noisy build and i have to understand why, but yeah, you can't access the drone build in the pr i linked but everything is green, very nice.. :)

drupol:
Hi @planctus,

To quote A. Einstein: 

> La connaissance s'acquiert par l'expérience, tout le reste n'est que de l'information.

I'm glad that such "experiment" leads to actually something worthy, it's a great example of collaboration between team, and also a good example of a true opensource spirit as well.

I just reviewed the PR and I notice that all the relative paths are gone, and I think that this is just a huge improvement.
Twig files are now closer to what they are supposed to be: following the PHP Twig standard.

However, I notice also that you removed the prefix, is it a good idea ?

What about names collisions with potentially existing templates ?
What is someone already has a `icon.html.twig` in its template then ?
Wouldn't it be better those prefixes them to avoid such issues ?

A couple of ideas to fix this and to discuss:
* Instead of registering all the paths, register only **one single path** (upper) and use: `ec-component/icon.html.twig`
* Use Twig namespaces like: `@ecl-ec/icon.html.twig` ?

Anyway, again, I'm glad to see these improvements and I can't wait to see this merged :-)

Have a good weekend all.

planctus:
hi @drupol, i added a namespace for the templates  in https://github.com/ec-europa/ecl-twig/pull/164, it's "@ecl-twig" ;)

drupol:
> hi @drupol, i added a namespace for the templates in #164, it's "@ecl-twig" ;)

Great :-)

If there are multiple "flavors" of ECL, what about having the flavor of ECL in the namespace ? Is it a good idea ?

planctus:
mmh..these "flavours" will only "affect" site header and footer, in theory, so that a dedicated namespace for them seems maybe too much. Anyway we are starting now introducing the "eu families" in the library as independent templates, we could even do like you suggested (like @ec-core/footer.html.twig, @ec-harmonised/footer.html.twig) but as these are going to be introduced with the maximum level of uncertainty about their evolution in the future, i would just name the files differently, keeping them in the same namespace. In any case it's a possibility, for sure, for now i'm trying to ensure that this new approach would be "supported" by all the people using this library at the moment.. ;)

planctus:
Can we close this @drupol ..? ;)

drupol:
Of course now that it has been merged ! Nice one :-)

PR #151, opened 2019-09-23
fix(site-header): Bridge language selector and overlay - TWIG-40

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +151,81 @@
...
+                    </ul>
+                  </div>
+                {% endfor %}
+              {% endif %}
+            </div>
+          </div>
+        </div>
       </div>
     </div>
     {% include '../ec-component-search-form/search-form.html.twig' with _search_form only %}
```
planctus: We don't  get the label of the input field, basing on the fix we just made on the search form i would expect the label to be set but hidden through its custom class.

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +151,81 @@
...
+          class="ecl-language-list ecl-language-list--overlay" 
+          aria-labelledby="ecl-language-list__title" 
+          role="dialog" 
+          data-ecl-language-list-overlay="true"
+        >
+          <div class="ecl-language-list__container ecl-container">
+            <div class="ecl-row">
+              <div class="ecl-language-list__close ecl-col-12 ecl-col-lg-8 ecl-offset-lg-2">
+               {% include '../ec-component-button/button.html.twig' with {
+                  label: _language_selector.overlay.close_label,
```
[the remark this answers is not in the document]
papegaill: Updated in the ec-component-button

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +151,81 @@
...
+              {% if _language_selector.overlay.items is not empty %}
+                {% for row in _language_selector.overlay.items|batch(_items_per_columns) %}
+                  <div class="ecl-language-list__column ecl-col-12 ecl-col-lg-4{{ loop.index == 1 ? ' ecl-offset-lg-2' }}">
+                    <ul class="ecl-language-list__list">
+                      {% for column in row %}
+                        <li class="ecl-language-list__item{{ column.is_active ? '  ecl-language-list__item--is-active' }}">
+                          {% include '../ec-component-link/link.html.twig' with { 
+                            link: { 
+                              type: 'standalone', 
+                              label: column.lang,
```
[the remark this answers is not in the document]
papegaill: solved by including the existing ec-component-language-list ;-)

src/ec/packages/ec-component-site-header/demo/data.js
```diff
@@ -0,0 +1,43 @@
...
+  adaptedData.logo.src = lng === 'en' ? englishBanner : frenchBanner;
+
+  adaptedData.language_selector = adaptedData.languageSelector;
+
+  adaptedData.language_selector.overlay.close_label =
+    adaptedData.languageSelector.overlay.closeLabel;
+
+  adaptedData.language_selector.overlay.items.forEach(item => {
+    if (item.isActive) {
+      item.is_active = true; // eslint-disable-line no-param-reassign
```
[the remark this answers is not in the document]
papegaill: done, thx for the tips ;-)

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -117,24 +112,24 @@
   <div class="ecl-site-header__container ecl-container">
     <div class="ecl-site-header__banner">
       <a 
-        class="ecl-link ecl-link--standalone" 
-        href="{{ _header_link.url }}" 
-        aria-label="{{ _header_link.aria_label }}"
+        class="ecl-link ecl-link--standalone"
+        href="{{ _logo.url }}"
```
planctus: we have "href", not "url" from the specs

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +138,20 @@
...
         </a>
+        {% include '../ec-component-language-list/language-list-overlay.html.twig' with { 
+          items: _language_selector.overlay.items,
+          close_label: _language_selector.overlay.close_label,
+          title: _language_selector.overlay.title,
+          icon_path: _icon_file_path,
+          extra_attributes: [
+            { name: 'aria-labelledby', value: 'ecl-language-list__title' }, 
+            { name: 'role', value: 'dialog' },
+            { name: 'hidden', value: '' }
```
planctus: I think we should improve the logic we have to set the extra_attributes, in ECL i see everytime these attributes that are not supposed to have any value but just to be set with an empty value "", we should be able to just set an attribute name (so we should check for the existence of the value key and only in that case print its value)

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +138,20 @@
                 name: 'language'
               }
             } only %}
-            <span class="ecl-site-header__language-code">{{ _language.code }}</span>
+            <span class="ecl-site-header__language-code">{{ _language_selector.code }}</span>
           </span>
         </a>
+        {% include '../ec-component-language-list/language-list-overlay.html.twig' with { 
+          items: _language_selector.overlay.items,
```
planctus: But this means that every implementation of this site header would need the developers to pass all the list of items in order to build the language list? I mean, this is a language switcher, it should ask you for the current url and for the rest it should work autonomously, i think, the language list is predefined as its order is, another parameter could be the type of language negotiation (suffix, prefix, underscore, slash, whatever...) and then it would build the list without needing to be fed, what do you think? :)

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -143,9 +138,20 @@
...
+          close_label: _language_selector.overlay.close_label,
+          title: _language_selector.overlay.title,
```
planctus: Why in ECL we get `/example#language_bg` as the href and here we have it empty..?

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -117,24 +112,24 @@
...
-          href="{{ _language.url }}"
-          data-ecl-language-selector
+          class="ecl-link ecl-link--standalone ecl-site-header__selector-link"
+          href="{{ _language_selector.url }}"
+          data-ecl-language-selector="true"
         >
-          {{- _language.label -}}
+          {{- _language_selector.name -}}
           <span class="ecl-site-header__language-icon">
             {% include '../ec-component-icon/icon.html.twig' with {
```
planctus: I see this attr in ECL for icons in the site-header `data-ecl-icon="true"`, not really sure it has any relevance, though.

src/ec/packages/ec-component-site-header/site-header.html.twig
```diff
@@ -117,24 +112,24 @@
...
+          src="{{ _logo.src }}"
         />
       </a>
       <div class="ecl-site-header__selector">
         <a 
-          class="ecl-link ecl-link--standalone" 
-          href="{{ _language.url }}"
-          data-ecl-language-selector
+          class="ecl-link ecl-link--standalone ecl-site-header__selector-link"
+          href="{{ _language_selector.url }}"
```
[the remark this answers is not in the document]
planctus: ok that we might have existing inconsistencies, but let's avoid introducing new ones. :)
At the top of the page, in the specs, you have this:
```
- "language_selector" (associative array) (default: predefined structure): Language switcher settings. format:
        {
          "href": (string) (default: ''): URL for switcher
```
So then i don't understand your concerns, yes we have to change it, both because like this we don't get any value from the specs we use from ecl and because we are saying: use "href" and then in the template we pass instead an unexisting url key.. ;)

src/ec/packages/ec-component-site-header/site-header.html.twig
planctus: Well, in this case we don't follow a suboptimal implementation in the ECL (remember that in the end that is just a demo...) Not asking you to fix it right now, (it's not enough not to pass the value), let's create a ticket for an improvement ... ;)

src/ec/packages/ec-component-site-header/site-header.html.twig
planctus: As agreed we will consider this for a future release.. ;)

src/ec/packages/ec-component-site-header/site-header.html.twig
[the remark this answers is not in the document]
papegaill: Sorry, I thought it was parameters for the language_selector items:
https://github.com/ec-europa/ecl-twig/pull/151/commits/e903b615b0c3d14d129dcbe208dd7f622dbcb797

Thanks, it's fixed now!

src/ec/packages/ec-component-site-header/site-header.html.twig
planctus: ok, apparently this is not relevant.

PR #180, opened 2019-10-17
feat(site-header-core): Add component - TWIG-79

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-site-header-core/site-header-core.html.twig
```diff
@@ -0,0 +1,190 @@
...
+      ]
+
+#}
+
+{# Default variables #}
+
+{% set _logo = logo|default({}) %}
+{% set _language_selector = language_selector|default({}) %}
+{% set _icon_file_path = icon_file_path|default('') %}
+{% set _auto_init = auto_init ?? true %}
```
papegaill: _auto_init parameter can be removed here, attribute is applied by default via the extra_attributes. To be checked in the site-header component also.

src/ec/packages/ec-component-site-header/__snapshots__/site-header.test.js.snap
```diff
@@ -96,6 +96,7 @@ exports[`EC - Site Header Default renders correctly 1`] = `
               </div>
               <div
                 class="ecl-language-list__title ecl-col-12 ecl-col-lg-8 ecl-offset-lg-2"
+                id="ecl-language-list__title"
```
papegaill: Updating site-header snapshots.. Should be site-header-core snapshots instead?

planctus: This is because the language list comes with an id that we didn't have, i added that in the component which is then included both by the site-header and site-header-core, you see this id now everywhere the language list overlay is used. You can check this markup on ECL:
https://ec.europa.eu/component-library/playground/ec/?path=/story/page-structure-languagelist--overlay

planctus: This is because the language list comes with an id that we didn't have, i added that in the component which is then included both by the site-header and site-header-core, you see this id now everywhere the language list overlay is used. You can check this markup on ECL:
https://ec.europa.eu/component-library/playground/ec/?path=/story/page-structure-languagelist--overlay

src/ec/packages/ec-component-site-header-core/site-header-core.html.twig
planctus: oh yeah...even though i think we should give the chance not to auto-init something, why do we force it? would be better to have it by default and leaving the chance to suppress it.

papegaill: This was approved in previous PR (https://github.com/ec-europa/ecl-twig/pull/149#discussion_r326595029) with @emeryro, but maybe we can see with platform team if they need this option?

src/ec/packages/ec-component-site-header/__snapshots__/site-header.test.js.snap
papegaill: ok sorry, I see that the snapshots for the site-header-core are indeed there.

PR #199, opened 2019-10-29
fix(expandable): use aria-expanded attribute - TWIG-103

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-expandable/expandable.html.twig
```diff
@@ -56,22 +56,14 @@
...
 
 <div class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  {% set _label = _label_expanded %}
   {% set _button_attributes = [
     { name: 'aria-controls', value: _aria_controls },
     { name: 'data-ecl-expandable-toggle', value: true },
     { name: 'data-ecl-label-expanded', value: _label_expanded },
-    { name: 'data-ecl-label-collapsed', value: _label_collapsed }
+    { name: 'data-ecl-label-collapsed', value: _label_collapsed },
+    { name: 'aria-expanded', value: _expanded }
```
planctus: the aria-expanded attribute is handled by javascript in this component, so we just set it initially as false (at least this is the way it is done in ECL) and then is js playing with that. I see that by doing like you did we could offer the possibility of using the component with a "not-collapsed" state initially, but this is not what the ECL definition of the component does, so i would not introduce it here on our side as an extra parameter.

src/ec/packages/ec-component-expandable/expandable.html.twig
```diff
@@ -83,7 +75,7 @@
   <div
     id="{{ _id ~ '-content' }}"
     class="ecl-expandable__content"
-    {{ _content_hidden_attribute }}
+    {% if not _expanded %}hidden{% endif %}
```
planctus: Same as before, this is handled by javascript, we set it as hidden and then js will play with that. I would have honestly considered this as an improvement in case it was meant to provide a non javascript solution with the collapsed button hidden and the content revealed, but it's not the case, so to stick to the definition of the component as we have it ECL we set this as hidden by default and then we leave the task to the js.

planctus:
@papegaill , i had a confirmation on the ECL side about this, personally speaking i would make this as an expandable/collapsible thing, but the collapsible "use case" is already covered by the accordion where we say not to use it with less than three pieces of information, so the single "collapsible" is not an use case, therefore no need to handle any logic on our side.

src/ec/packages/ec-component-expandable/expandable.html.twig
papegaill: Agree but this was already the case before the actual PR, I was also a bit confused with this, see the parameter here:
https://github.com/ec-europa/ecl-twig/blob/292c833c54b31c19edfdd97042fb16dc5cb405a5/src/ec/packages/ec-component-expandable/expandable.html.twig#L8

Anyway, I have removed the `expanded` parameter now.

src/ec/packages/ec-component-expandable/expandable.html.twig
papegaill: Same as my previous comment this was already implemented... See: https://github.com/ec-europa/ecl-twig/blob/292c833c54b31c19edfdd97042fb16dc5cb405a5/src/ec/packages/ec-component-expandable/expandable.html.twig#L66

Updated now.

PR #270, opened 2019-12-17
feat(ecl-diff): Adding command to diff a component against the ECL markup - TWIG-180

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


php/php_storybook/scripts/ecl-diff.js
```diff
@@ -0,0 +1,312 @@
...
+    type: 'list',
+    describe: 'The group the component is associated with in ECL',
+    choices: ['components', 'page-structure'],
+  },
+  eclSubSection: {
+    type: 'list',
+    describe: 'Is the component nested into a sub-section?',
+    default: 'none',
+    prompt: 'always',
+    choices: ['none', 'forms', 'navigation', 'banners'],
```
papegaill: I think you should add 'footers', 'site headers', 'page headers', 'banners'?

planctus: Why @papegaill? banners is already in place, site-headers and page-headers is not anything in the "path" of the ecl story, the site-header-core for instance you reach it at: components-site-headers-core--default, so there is no sub-section for them. ( the same for page headers) ;)

php/php_storybook/scripts/ecl-diff.js
```diff
@@ -0,0 +1,312 @@
...
+  interactive: { default: true },
+  component: {
+    type: 'list',
+    choices: packages,
+    describe: 'The component you want to diff',
+  },
+  variant: {
+    type: 'input',
+    default: '',
+    describe: 'The name of the twig variant (what is after the -- Ex: audio) ',
```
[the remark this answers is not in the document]
planctus: eh, no...many of our rendered files are simply having the name of the component, it totally depends on the name of the spec file and many times they don't set "default", that's why is like this.. ;)

PR #284, opened 2020-01-14
feat(site-header-harmonised): Adding group3 - TWIG-200

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.story.js.bckp
```diff
@@ -45,4 +47,19 @@ storiesOf('Components/Site Headers/Harmonised', module)
...
+      siteHeaderHarmonised(
+        merge(dataGroup3, {
+          logo: {
+            src:
+              'https://inno-ecl.s3.amazonaws.com/media/examples/placeholder.svg',
+          },
+        })
+      ),
+    {
+      notes: { markdown: notes },
```
planctus: I think we can have the "hardcoded" notes replaced by the real data used to render the component as we do for the other two groups, no..? I refer to the additional parameter in the notes object, this is what we use to pass the real data to the storybook panel.

src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.story.js.bckp
```diff
@@ -45,4 +47,19 @@ storiesOf('Components/Site Headers/Harmonised', module)
...
     }
+  )
+  .add(
+    'group 3',
+    () =>
+      siteHeaderHarmonised(
+        merge(dataGroup3, {
+          logo: {
+            src:
+              'https://inno-ecl.s3.amazonaws.com/media/examples/placeholder.svg',
```
planctus: This should come from the adapter, the things we do in the stories are not retrieved by php-storybook (the one you start with start:php-ec), so better working in the specs or in the adapter when it's possible.

src/ec/packages/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig
```diff
@@ -156,7 +156,7 @@
 {% endif %}
 
 {# Print the result #}
-
+{% if group != 'group3' %}
```
planctus: Mmhh, despite the fact that the template produces the right html (perfectly) i don't necessarily understand the reason why duplicating so much code, it looks like this "variant" is just a very simplified version of the other two, but it's markup corresponds to the "first part" of the template, so i would maybe use the same. If you did this on purpose considering also this other option don't consider this as a "mandatory" thing to be changed, you can explain the reasons why you chose to do like this.. ;)

src/ec/packages/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig
```diff
@@ -171,125 +171,138 @@
...
-            <hr className="ecl-site-header-harmonised__login-separator" />
-            {% include '@ecl-twig/ec-component-link/ecl-link.html.twig' with {
-              link: {
-                label: _login_box.label,
-                path: _login_box.href,
-                type: "standalone"
-              }
-            } only %}
-          </div>
+        {% if group != 'group3' %}
```
planctus: We have a local variable _group, please use that cause it defines a default and the value for the group parameter might be simply missing.

src/ec/packages/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig
```diff
@@ -171,125 +171,138 @@
...
-                type: 'general',
-                name: 'log-in',
-                size: 'm'
-              },
-              extra_classes: 'ecl-site-header-harmonised__icon'
-            } only %}
-            {{- login_toggle.label_not_logged -}}
-          </a>
+        </a>
+        {% if site_name is not empty %}
```
planctus: MMhh..yeah, i haven't noticed there was this difference in the two markups when i told you to re-use the existing section of the template, i'd say that to improve it a bit we could define this chunk of html as a variable when we process the variables and then print the variable, just to avoid the duplicated code.
`<div class="ecl-site-header-harmonised__site-name">{{ site_name }}</div>`

src/ec/packages/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig
```diff
@@ -171,125 +171,138 @@
...
       {% endif %}
-      {% if _language_selector is not empty %}
-        <a
-          class="ecl-link ecl-link--standalone ecl-site-header-harmonised__language-selector"
-          href="{{ _language_selector.href }}"
-          data-ecl-language-selector
-        {% if _has_overlay is defined %}
-          aria-controls="language-list-overlay"
-          aria-expanded="false"
+
```
planctus: General issue in the template now, alignment, indentation. I know it's hard to keep the indentation especially when we use logic in the templates, but we should try to preserve the indentation of the html tags, a closing tag has to correspond to the opening tag, we have two spaces indentation for all the tags, if(s) can be adapted, indentation of the html tags no.

src/ec/packages/ec-component-site-header-harmonised/demo/data--group3.js
```diff
@@ -0,0 +1,12 @@
+import specs from '@ecl/ec-specs-site-header-harmonised/demo/data--group3';
```
planctus: i remember we talked about an image that you were setting in the story to be moved here, wasn't needed, that's the reason why i don't see it here?

src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.story.js.bckp
```diff
@@ -45,4 +47,14 @@ storiesOf('Components/Site Headers/Harmonised', module)
     {
       notes: { markdown: notes, json: dataGroup2 },
     }
+  )
+  .add(
+    'group 3',
+    () =>
+      siteHeaderHarmonised(
+        merge(dataGroup3, {})
```
planctus: What is this..? :)
Why merging the specs with an empty object? We use the merge when we want to pass additional data or we want to override some of the existing, in this case i see we don't need anything, instead.

src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.story.js.bckp
```diff
@@ -8,6 +8,8 @@ import logo from '@ecl/ec-resources-logo/logo--en.svg';
 import siteHeaderHarmonised from './ecl-site-header-harmonised.html.twig';
 import dataGroup1 from './demo/data--group1';
 import dataGroup2 from './demo/data--group2';
+import dataGroup3 from './demo/data--group3';
```
planctus: The file has issues with the linter, they are not appearing now because the file is renamed and not being checked, but what you should do is to rename it in story,js, run` yarn eslint --fix pathToTheFile` and then rename it again in .bckp and push it.

planctus:
Fine for me, let's wait Filipa's for the "approval" ;)

src/ec/packages/ec-component-site-header-harmonised/__snapshots__/site-header-harmonised.test.js.snap
```diff
@@ -11661,3 +11661,104 @@ exports[`EC - Site Header Harmonised Group 2 renders correctly with extra class
   </div>
 </header>
 `;
+
+exports[`EC - Site Header Harmonised Group 3 renders correctly 1`] = `
+<header
+  class="ecl-site-header-harmonised ecl-site-header-harmonised--group3"
+  data-ecl-auto-init="SiteHeaderHarmonised"
```
planctus: This attribute is not set for the group3 "variant" in ECL

src/ec/packages/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig
planctus: I think in the end we just made a lot of confusion around this, @Joosthe, everything was based on the fact that you had duplicated code, now that this is a one time markup that you set, i don't see anymore the value of storing this html in a variable.

src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.test.js
```diff
@@ -40,7 +41,7 @@ describe('EC - Site Header Harmonised', () => {
   });
 
   describe('Group 2', () => {
-    dataGroup1.group = 'group2';
+    dataGroup2.group = 'group2';
```
planctus: Ok, this should come from the specs, because if you had to do it here it means that is not there. I see you did it for the group3 the correct way, you should also align the other two groups (well, group1 is the default...so), group2 for instance is set in the story and if you check the rendered markup `yarn check:component site-header-harmonised` you will notice that the class is wrong for group2, cause it gets the default.

src/ec/packages/ec-component-site-header-harmonised/site-header-harmonised.test.js
```diff
@@ -69,4 +70,34 @@ describe('EC - Site Header Harmonised', () => {
       return expect(render(withExtraAttributes)).resolves.toMatchSnapshot();
     });
   });
+
+  describe('Group 3', () => {
+    dataGroup3.group = 'group3';
```
planctus: This already comes like this from the specs, why are you setting it again?

PR #291, opened 2020-01-17
fix(button): Setting data-ecl-icon on the icon when used in a button - TWIG-199 

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
```diff
@@ -54,23 +54,35 @@
...
+          <span class="ecl-button__container">
+            {% include '@ecl-twig/ec-component-icon/ecl-icon.html.twig' with {
+              icon: {
+                path: _item.toggle.icon.path,
+                type: "general",
+                name: _item.toggle.icon.name,
+                size: "s"
+              },
+              extra_classes: 'ecl-icon ecl-icon--s ecl-button__icon ecl-button__icon--before',
+              extra_attributes: [{ name: "data-ecl-icon" }]
```
planctus: I'm seeing this attribute also in other components including an icon, are we sure this is not a default attribute of this component, now?

src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
```diff
@@ -54,23 +54,35 @@
...
+            </span>
+          </span>
+        </button>
+        
       </h{{ _item.level|default(3) }}>
       <div
         class="ecl-accordion__content"
         hidden
-        id="{{ _item.id }}"
+        id="{{ _item.id }}-content"
```
planctus: Not really sure why this is done like this in ECL, might be that this a way to "target" the old accordion (this is the deprecated one).

src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
```diff
@@ -54,23 +54,35 @@
...
-          type: 'button',
-          icon: _item.toggle.icon,
-          icon_position: 'before',
-          extra_classes: 'ecl-accordion__toggle',
-          extra_attributes: [
-            { name: 'aria-controls', value: _item.id },
-            { name: 'data-ecl-accordion-toggle' },
-          ]
-        } only %}
+        <button
```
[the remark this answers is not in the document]
Joosthe: The issue here was to give the a give and attribute specifically for the accordion. I have discussed this with @papegaill and we thought this was the better way since this is also used in the accordion2

src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
Joosthe: Maybe, not in the ecl librarie the data attribute isn't defined either on the standard icon. It is just set  on certain icons. I don't know if we should fix this is both libraries or just leave it like this.

src/ec/packages/ec-component-button/ecl-button.html.twig
```diff
@@ -61,14 +61,20 @@
       {% if _icon_position == 'before' %}
         {% include '@ecl-twig/ec-component-icon/ecl-icon.html.twig' with {
           icon: _icon,
-          extra_classes: 'ecl-button__icon ecl-button__icon--before'
+          extra_classes: 'ecl-button__icon ecl-button__icon--before',
+          extra_attributes: [
+            { name: 'data-ecl-icon', value: 'true' }
```
planctus: So....our approach with boolean attributes follows the html 5 defintion:
boolean attributes should be defined by their name only, setting a "true" value doesn't make them boolean, true is a string and unfortunately we see these custom booleans in ECL always with a quoted value which is basically wrong, so we don't have to reproduce that.
The only use case where we would set the "true" value as a string is when it is used in javascript (which should be avoided).
The alternative would be to set the true value unquoted in order for it to be a boolean, but it's confusing, that is also why we try to handle these differences with ECL when we make a diff.

src/ec/packages/ec-component-button/ecl-button.html.twig
```diff
@@ -61,14 +61,20 @@
...
         } only %}
       {% endif %}
       <span class="ecl-button__label" data-ecl-label="true">{% block label _label|raw %}</span>
       {% if _icon_position == 'after' %}
         {% include '@ecl-twig/ec-component-icon/ecl-icon.html.twig' with {
           icon: _icon,
-          extra_classes: 'ecl-button__icon ecl-button__icon--after'
+          extra_classes: 'ecl-button__icon ecl-button__icon--after',
+          extra_attributes: [
+            { name: 'data-ecl-icon', value: 'true' }
```
planctus: same as before

src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
```diff
@@ -66,6 +66,7 @@
             { name: 'data-ecl-accordion-toggle' },
           ]
         } only %}
+
```
planctus: but i remember you were adding the "-content" to the ids like in ECL in the accordion, why did we remove that? (mine was just a comment, it is a bit weird but as i wrote maybe there was a reason behind that.

PR #297, opened 2020-01-23
feat(select): (almost) green diff with ECL - TWIG-240

planctus:
# PR description

The reason why the diff is not perfectly "green" is that the ECL react component is missing the required attribute when the field is required.
A ticket has been created to handle this on the ECL side:
https://webgate.ec.europa.eu/CITnet/jira/browse/FRONT-11

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-select/select.story.js
```diff
@@ -36,11 +36,10 @@ storiesOf('Components/Forms/Select', module)
         optional_text: text('Optional text', '(optional)'),
         width: inputWidthSelect,
         id: 'example-id',
```
papegaill: necessary to set and example id here?

planctus: Right, but more than that i saw that we were not using the data from the local folder but still the specs, i improved the story file, now it should be fine ;)

PR #300, opened 2020-01-23
feat(text-area): making diff match - TWIG-241

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-text-area/ecl-text-area.html.twig
```diff
@@ -110,7 +110,9 @@
     id="{{ _id }}"
   {% endif %}
     class="{{ _css_class }}"
+  {% if _name and _name is not empty %}
```
planctus: _name is a local variable that we define at the top of the template, but its default is an empty string, so here you only need to check for "is not empty" ;)

src/ec/packages/ec-component-text-area/adapter.js
```diff
@@ -13,10 +13,14 @@ const adapter = initialData => {
...
-  adaptedData.helper_text = "This is the input's helper text";
+  adaptedData.helper_text = "This is the input's helper text.";
   adaptedData.optional_text = '(optional)';
-  adaptedData.name = 'example-name';
   adaptedData.label = 'Label';
+  adaptedData.required = true;
+  adaptedData.required_text = '*';
+  adaptedData.optional_text = '(optional)';
+  adaptedData.id = 'example';
+  adaptedData.name = '';
```
planctus: We don't really need to add the key if we are not setting any value, this is not set in the specs so there's nothing to override.

src/ec/packages/ec-component-text-area/adapter.js
```diff
@@ -13,10 +13,14 @@ const adapter = initialData => {
...
 
-  adaptedData.helper_text = "This is the input's helper text";
+  adaptedData.helper_text = "This is the input's helper text.";
   adaptedData.optional_text = '(optional)';
-  adaptedData.name = 'example-name';
   adaptedData.label = 'Label';
+  adaptedData.required = true;
+  adaptedData.required_text = '*';
+  adaptedData.optional_text = '(optional)';
+  adaptedData.id = 'example';
```
planctus: Why if we set it here we don't use this in the story, then..? (i know it's weird that we have ids in the specs but then we need to override them to make the diff match).

Joosthe: Keeping this id for the story also fine.I just overridden the story because i thought it the story itself need to match the previous state.

PR #305, opened 2020-01-24
feat(table): Extra attributes and extra classes for the table rows - TWIG-203

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-table/ecl-table.html.twig
```diff
@@ -10,16 +10,18 @@
...
+        "extra_classes": (string) (default: ''),
         "label" (string or array of string)
         "data-ecl-table-header" (string) (default: ''),
         "group" (bool) (default: false),
       },
       ...
     ]
   - "extra_classes" (string) (default: '')
-  - "extra_attributes" (array) (default: []): format: [
+  - "extra_attributes" (string) (default: []): format: [
```
papegaill: Main extra_attributes should stay as an array? Also, for rows, why not keeping same array structure, how to do if you want to pass an extra_attribute with a value?

planctus: You're right about the fact that i also changed the casting for the main "extra_attributes" and that is unwanted. But for the "new" extra_attributes for the rows, in order to use the same approach we have for the main attributes we should also apply the same logic to them, if it comes as an array we have to process it and build a string out of it. This is also happening in a for loop in this case, so putting all this processing there looks really overwhelming, if you have the value you pass the full attribute (name="value"), anyway.. ;)

PR #307, opened 2020-01-27
feat(timeline): Green diff with ECL - TWIG-244

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-timeline/ecl-timeline.html.twig
```diff
@@ -69,7 +69,7 @@ Parameters:
     {% for _item in _items %}
       {# Decide on per item if it should display. #}
       {% set extra_class = (loop.index > _display_amount and loop.revindex > _bottom_amount) ? ' ecl-timeline2__item--collapsed' : '' %}
-      <li class="ecl-timeline2__item{{ extra_class }}" id="{{ _item.id }}">
+      <li class="ecl-timeline2__item{{ extra_class }}" {% if item.id is not empty %} id="{{ _item.id }}" {% endif %}>
```
planctus: @Joosthe, everytime we define our local variables out of the input data then we should use that. Here you are checking item.id and then echoing _item.id.

planctus: But, @Joosthe, you forgot about this? You are looping and your current element in the loop is _item, why are you checking for item.id? I also see that there is no evidence that the id property will be in place, so the check has to be complete (`if _item.id is defined and _item.id is not empty`) (here is where you have the empty spaces at the end of two lines..)

src/ec/packages/ec-component-timeline/ecl-timeline.html.twig
```diff
@@ -69,7 +69,7 @@ Parameters:
...
         <div class="ecl-timeline2__label">{{ _item.label }}</div>
         {% if _item.title is not empty %}
```
planctus: Also this is not anything you touched, but i don't see the reason why we are assuming that _item.title is going to be in place, twig is not php, !empty doesn't mean "if isset and is not empty", we need to check also for the property itself, if `_item.title is set and _item.title is not empty`

PR #313, opened 2020-01-29
feat(php-storybook): Integration with the CI environment

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


scripts/netlify-switch.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
kalinchernev: `netlify.toml` is an optional file in the setup and you will most probably be able to achieve the same simply deleting it and passing arguments to the specific tools.

For instance, the build command is part of the npm script and you can call it directly from [drone's commands](https://github.com/ec-europa/ecl-twig/pull/313/files#diff-3216dfff0ed3e301453e6799e8c367e2R56) and the publishing directory is a flag in the [deploy command of the cli](https://cli.netlify.com/commands/deploy) so it's possible to pass it in [this command](https://github.com/ec-europa/ecl-twig/pull/313/files#diff-3216dfff0ed3e301453e6799e8c367e2R63).

scripts/set-netlify-deployment-status.js
```diff
@@ -31,6 +32,10 @@ const run = async () => {
     return;
   }
 
+  const contextName = !DEPLOY_CONTEXT
```
kalinchernev: Can be named `context` to have an object [value shorthand](https://alligator.io/js/object-property-shorthand-es6/) later like 

```
payload = {
        state: 'success',
        target_url: deploymentResult.deploy_url,
        description: 'Preview ready!',
        context,
      };
```

Feel free also to break out of `drone/netlify` naming as it was quite a brisk initial one around the preview feature.

Maybe something like `drone/twig-js` and `drone/twig-php` will be more descriptive in your case to compare 2 previews related to rendering options.

planctus:
Thanks @kalinchernev, think all of your recommendations were good, i basically applied what you suggested me and everything is working fine ;)

kalinchernev:
@planctus nice, glad to hear it works better
I think you can further tweak the drone configuration to save some time.
The https://github.com/ec-europa/ecl-twig/blob/398d46789c4e37947af77fa44b84162c04984486/.drone.yml#L26 `deploy-preview` and `php-storybook` are not dependent to each other as far as I can tell. This means you freely put them next to each other like `deploy-twigjs-preview` and `deploy-twigphp-preview` and add a `group: previews` to them to run them in parallel. There's a small chance the netlify service times out, but I doubt it.

Another small thing which i noticed and have experienced in ecl repo is the case of failing deployment. In ecl, the deployment result is no longer stored in file, because the whole process information is lost because of the piping. Instead, the info is taken separately so that you can see if there's been a failure on neltify in the CI itself. https://github.com/ec-europa/europa-component-library/pull/1423

These are only suggestions of course

planctus:
I've tried already @kalinchernev to group the two previews, but i had errors: https://drone.fpfis.eu/ec-europa/ecl-twig/2165, i'll take a look at the other suggestion.. ;)

PR #318, opened 2020-01-30
feat(card): Improve knobs on component- TWIG-254

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-card/card.story.js
```diff
@@ -2,62 +2,205 @@
...
+            },
+            description: text('Description', dataCard.card.description),
+            image: {
+              src: text('Image path', dataCard.card.image.src),
+            },
+            meta: array('Metatags', dataCardMeta, '|'),
+            infos: [
+              {
+                icon: {
+                  size: 'xs',
```
planctus: I would avoid "hardcoding" these information here in the story, we should rely on the specs. I understand that the use case of a merge here is not the easiest ever but as said in the previous comment there is no need for an overall merge if that is overcomplicating things, what we want to have is knobs using default values coming from the specs, basically.

src/ec/packages/ec-component-card/card.story.js
```diff
@@ -2,62 +2,205 @@
...
+const dataCardMeta = [];
+const dataCardTags = [];
+const dataTileLinks = [];
+const dataCardTagMeta = [];
+const dataCardTagTag = [];
+const dataCardEventInfos = [];
 
-  return data;
-};
+prepareMerge(dataCard.card.infos, dataCardInfos, 'label');
```
planctus: It took me a while to track back all the reasoning that led you to this implementation, it feels to me that we are overdoing here, basically all of this is done for then merging the specs with keys and values that we almost entirely set in the story file.
So that all this processing would just not be needed in case we would simply pass the object to the story without merging it with the specs.

Tell me if this makes sense to you or if it's totally unclear, anyway in the code you currently have duplicated operations apparently.
What you do in the forEach would be normally done using map in javascript: https://www.w3schools.com/jsref/jsref_map.asp 
and the parameters name you used are very confusing cause you actually have two arrays but you name them "..ArrayName"

In my opinion we could reduce the amount of code and processing of these "small" amount of data and we should not set in the story "values" that we receive from the specs, but the knobs are working fine and i know you already worked a lot on this, so i'll let you decide whether to just "fix" what really needs to be fixed or to look for a simpler way to achieve the same result. ;)

src/ec/packages/ec-component-card/card.story.js
```diff
@@ -25,9 +26,28 @@ const formatInfo = data => {
 storiesOf('Components/Card', module)
   .addDecorator(withCode)
   .addDecorator(withNotes)
+  .addDecorator(withKnobs)
   .add(
     'card',
     () => {
+      dataCard.card.image.src = text('Image', dataCard.card.image.src);
```
planctus: @Joosthe , at this point my question is: do we really want to find a way to overcomplicate things? :D
I understand that this works, but why did you decide to go this way instead of keeping things as they were just removing the "merge" that was the reason of the previous approach overheads? I mean, i really don't want to stress you, but i find this completely unreadable, while i was simply expecting you to create the card object (as it was before, as a key (string) value (from the specs) in the story, with the knob when you needed it, just without merging any pre-existing object from the specs because totally unneeded and with problematic consequences.
please tell me if there is anything unclear about this, we've been revolving this "improvement" enough, i think :)

PR #319, opened 2020-01-31
feat(knobs): unwantend knobs on first load - FRONT-306

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


planctus:
@Joosthe , think there is still something odd, didn't look at the code but i see this, should be something trivial to fix:
image

Knobs are working fine and the "bug" is fixed, anyway ;)

planctus:
the size of the icon change in the page-header, it should be "s" and it's "m" (in the test), i see that you remove the "merge" of that parameter in the component, but the reason for that is that we wanted to force this value, not letting the user change it, basically. Please out it back ;)

planctus:
In general, not that it's always the case, we tend to avoid for loops, forEach for instance might be easier to implement and to read.

PR #322, opened 2020-02-04
feat(knobs): update fact-figures component - FRONT-450

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-fact-figures/ecl-fact-figures.html.twig
```diff
@@ -54,7 +54,7 @@
   <div class="ecl-fact-figures__items">
   {% for item in items %}
     <div class="ecl-fact-figures__item">
-    {% if item.icon is defined and item.icon is not empty %}
+    {% if item.icon is defined and item.icon is not empty and display_icons == 1 %}
```
planctus: This is a boolean, you can check for it's true value simply like this: `and display_icons`, @Joosthe

src/ec/packages/ec-component-fact-figures/ecl-fact-figures.html.twig
```diff
@@ -80,7 +80,7 @@
   </div>
 {% endif %}
 
-{% if view_all.link is defined and view_all.link is not empty %}
+{% if view_all.link is defined and view_all.link is not empty and view_all.visible == 1  %}
```
planctus: like before

planctus: Also in this case i would first create a local variable:
[{% set _view_all_visibility = view_all.visible|default(false) %}](url)
and then you would check for `and _view_all.visibility`

src/ec/packages/ec-component-fact-figures/ecl-fact-figures.html.twig
Joosthe: Indeed I totally forgot to update readme and the template + i have set the code to local variables

src/ec/packages/ec-component-fact-figures/ecl-fact-figures.html.twig
planctus: There is a double space at the end of line 83, this means that you still don't have grumphp running when you commit, @Joosthe , can you please verify this? Otherwise we would have this problem everytime.

planctus: On line 34 there are two spaces at the end of the line, also

papegaill: I don’t really understand the usefulness of this _view_all.visible_ parameter. if the link should not appear, I think that we just have to remove the view_all object?

planctus: Indeed i agree that is redundant, it's not entirely clear the reason why one would define this view_all object if the intention is not to show the link it generates.
I thought the same parameter existed in ECL and that's why we were adding it on our side, but in ECL it doesn't exist, so i agree about simply removing it.

papegaill: ok I'll update, thx

PR #334, opened 2020-02-11
feat(tag): Improve knobs on component - FRONT-492

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -11,6 +11,12 @@ import dataRemovable from './demo/data--removable';
 import tag from './ecl-tag.html.twig';
 import notes from './README.md';
 
+const variants = {
+  'As a link': 'link',
+  'As a Button': 'button',
+  Removable: 'removable',
```
papegaill: missing quotes to _Removable_?

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -23,7 +29,9 @@ storiesOf('Components/Tag', module)
           tag: {
             label: text('Label', 'Link tag'),
             path: text('Url', '/example'),
+            type: select('Variable', variants, 'link'),
```
papegaill: What's the purpose of having variants types as knobs if it's already defined as stories?

Joosthe: This was literally the scope of the ticket to build in this functionality.

planctus: Hi @papegaill , thanks for the review, your concern is clearly reasonable, this new select will let you switch between the different variants you have for the component which are indeed matching the stories we demo.
But if i have to consider what is redundant here, i'd say it's the "multiplication" of the stories for something that is actually just a parameter switch, not the availability of the parameter really handling the switch.
At the same time, i see that sometimes stories are a more evident way to show the different "variants" available per component, but we have to make it clear what is the real difference between them.
The label should be "variant", though, for this to happen, "variable" is not really identifying what the parameter is about

papegaill: @planctus I think it's confusing, because for example if on the "_as a link_" story i switch variant, I'm then able to set "_url_" on a button or a removable tag. same with parameter "_aria-label_" available only on removable tag.

planctus: ok, you convinced me, let's remove it ;)

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -21,9 +21,10 @@ storiesOf('Components/Tag', module)
       tag(
         merge(dataLink, {
           tag: {
-            label: text('Label', 'Link tag'),
-            path: text('Url', '/example'),
+            label: text('Label', dataLink.tag.label),
+            path: text('Url', dataLink.tag.path),
```
planctus: MMhh, but if we are doing this in order to match as much as possible the specs, why we name the knob "url" when the property passed to the template is path? I understand that in any case a bit of ambiguity is in place, considering that both a path and an url can be used, so that's what i would do, we name it: `Path (accepts also a full url)`, what do you think?

PR #344, opened 2020-02-20
feat(demo): Footer core - Adding mandatory elements knobs - FRONT-830

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-footer-core/footer-core.story.js
+++ b/src/ec/packages/ec-component-footer-core/footer-core.story.js
@@ -1,11 +1,10 @@
+/* eslint-disable no-param-reassign, no-shadow */
 import { storiesOf } from '@storybook/html';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
 import withCode from '@ecl-twig/storybook-addon-code';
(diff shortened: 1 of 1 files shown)


src/ec/packages/ec-component-footer-core/footer-core.story.js
```diff
@@ -1,14 +1,109 @@
...
+      'sections[0].title.link.label',
+      data[0].title.link.label,
+      requiredGroupId
+    ),
+    path: text(
+      'sections[0].title.link.path',
+      data[0].title.link.path,
+      requiredGroupId
+    ),
+    placeholder: text('optional elements', 'none', optionalGroupId),
```
papegaill: Strange? I think it's confusing, maybe better to have the tabs and nothing inside it?

planctus: I agree, but the tabs would simply not appear in case there is only one group defined.
That's the reason why of that "fake" prop, but i don't like it either, just that i think it's important to show the "mandatory elements" tab, since this is not going to be a single use case we can try finding a better solution, nothing against that, for now i think i will change the text inside the field to be something like "no optional element in this story"

PR #346, opened 2020-02-27
feat(knobs): Blockquote panel with three sub-panels, buttons for show/hiding - FRONT-803

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,17 +9,49 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
+};
+
+const formatQuote = passeddata => {
+  // declare your knobs when you format your data
+  const citation = passeddata.citation
+    ? text('Citation', passeddata.citation, requiredGroupId)
+    : false;
+
+  const author = passeddata.author
+    ? text('Author', passeddata.author, optionalGroupId)
```
planctus: @Joosthe . where did you see that the author  is an optional parameter?
https://ec.europa.eu/component-library/ec/components/blockquote/usage/
Due to this also the check you added in the template i think is not ok, this falls into the other track about enforcing some check on the input data.

src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,17 +9,49 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
   .addDecorator(withNotes)
   .add(
     'default',
-    () =>
-      blockquote({
-        author: text('Author name', data.author),
-        citation: text('Citation', data.citation),
-      }),
+    () => {
+      button('show/ hide author', authorBtnToggler, useCasesGroup);
```
planctus: I think we should try to be consistent in labelling these buttons, i'm not that happy with the "with or without" but if we want to change it we have to do it consistently. Anyway, this button in particular has to be removed, the author is not an optional element, as seen before.

src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,17 +9,49 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
+
+      return blockquote(formatQuote(defaultData));
+    },
     {
       notes: { markdown: notes, json: data },
```
planctus: You are passing to the note panel the data as it come from the specs (or the adapter), not the one we really pass to the template, therefore it never changes even if you hide the author. (You will remove this button so the problem will not be in place, but in this implementation you can see that the note panel is neither updated when you change the data using the button nor when you use the knobs ;) )

src/ec/packages/ec-component-blockquote/ecl-blockquote.html.twig
```diff
@@ -40,7 +40,9 @@
 
 <blockquote class="{{ _css_class }}"{{ _extra_attributes|raw }}>
   <p class="ecl-blockquote__body">{{ _citation }}</p>
+  {% if _author  %}
```
planctus: As mentioned before, being this a mandatory element there is not reason for this check, we have to enforce some check on these mandatory elements but not this way.

src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,17 +9,49 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
+
+// Preserve the adapted specs.
+const defaultData = { ...data };
+
+// declare your toggled data
+const authorBtnToggler = () => {
+  defaultData.author = defaultData.author ? false : data.author;
+};
+
+const formatQuote = passeddata => {
```
planctus: passeddata? :) something different, maybe..? 
I wonder the reason why you are checking for the passeddata.citation like it might not be in place.
In your logic the author might not be present, but as we saw also author is a mandatory field, so...

src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,18 +9,48 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
+  // declare your knobs when you format your data
+  const citation = prepareData.citation
+    ? text('Citation', prepareData.citation, requiredGroupId)
+    : false;
+
+  const author = prepareData.author
+    ? text('Author', prepareData.author, requiredGroupId)
+    : false;
+
+  const newQuote = {
```
planctus: I do see this approach as a bit redundant, why do you have to build to separated constants, redefining entirely the object to pass to the template?
The idea is that you pass the data to this function and it should return this data formatted, in this case where don't have any button, it is straightforward.
You basically receive the data and inside of it you are only going to "override" what you need and return it. I'm attaching here how i would make the story file, up to you to use it as such or make it your own.. ;)
```suggestion
/* eslint-disable no-param-reassign, no-shadow */
import { storiesOf } from '@storybook/html';
import { withKnobs, text } from '@storybook/addon-knobs';
import { withNotes } from '@ecl-twig/storybook-addon-notes';
import withCode from '@ecl-twig/storybook-addon-code';

import data from '@ecl/ec-specs-blockquote/demo/data';

import blockquote from './ecl-blockquote.html.twig';
import notes from './README.md';

// Labels for the group ids
const optionalGroupId = 'Optional elements';
const requiredGroupId = 'Mandatory elements';

const prepareQuote = data => {
  data.citation = text('citation', data.citation, requiredGroupId);
  data.author = text('author', data.author, requiredGroupId);

  return data;
};

storiesOf('Components/Blockquote', module)
  .addDecorator(withKnobs)
  .addDecorator(withCode)
  .addDecorator(withNotes)
  .add(
    'default',
    () => {
      const storyData = prepareQuote(data);
      storyData.demo = text(
        'optional elements',
        'No optional element in this story',
        optionalGroupId
      );

      return blockquote(prepareQuote(storyData));
    },
    {
      notes: { markdown: notes, json: data },
    }
  );
```

src/ec/packages/ec-component-blockquote/blockquote.story.js
```diff
@@ -8,18 +9,48 @@ import data from '@ecl/ec-specs-blockquote/demo/data';
...
-        citation: text('Citation', data.citation),
-      }),
+    () => {
+      button('fake button to enforce groups', authorBtnToggler, useCasesGroup);
+
+      return blockquote(prepareQuote(defaultData));
+    },
     {
-      notes: { markdown: notes, json: data },
+      notes: { markdown: notes, json: defaultData },
```
planctus: In general i would avoid passing the data with the knobs to the note panel, the logic is that we should have already good data to pass to it as they come form the adapter, then we use the "formatting" function inside the story file only to set the knobs we want and override values for icons and logos, basically. what i fear is that if we pass the knobs this will "call" the knobs again with an unchanged story...This might be already the case in some other component, not sure, but would be better to avoid.

planctus:
@Joosthe , left a few comment and a suggestion for the story file, tell me if there is anything unclear or "debatable" .. ;)

src/ec/packages/ec-component-blockquote/blockquote.story.js
planctus: @Joosthe , i would really like to close this, i gave you this suggestion but you did it a bit differently, ok for me, but:
now you have the "demo" parameter in the notes tab, which should not be there and in my suggestion would not have appeared. 
This is  because you are still "cloning" data into another variable, (with a typo: "preperaData") and the demo property you set on the storyData var get assigned also to the main data object.

PR #347, opened 2020-02-27
feat(knobs): Button show/hiding elements and grouping knobs - FRONT-865

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-button/button.story.js
```diff
@@ -45,86 +83,114 @@ storiesOf('Components/Button', module)
...
-          icon_position: select('Icon position', iconPositionSettings, 'after'),
-          disabled: boolean('Disabled', false),
-        })
-      ),
+    () => {
+      return button(prepareButton(primaryData));
+    },
     {
-      notes: { markdown: notes, json: prepareButton(dataPrimary) },
+      notes: { markdown: notes, json: { button: primaryData } },
```
planctus: i don't see the "demo" input data in the notes being updated after touching the knobs.

src/ec/packages/ec-component-button/button.story.js
```diff
@@ -45,86 +83,114 @@ storiesOf('Components/Button', module)
...
-      ),
+    () => {
+      return button(prepareButton(searchData));
+    },
     {
-      notes: { markdown: notes, json: prepareButton(dataSearch) },
+      notes: { markdown: notes, json: { button: searchData } },
     }
   );
+// .add(
```
planctus: You're keeping this commented code here on purpose?

src/ec/packages/ec-component-button/button.story.js
planctus: but it's still not good, @Joosthe 
image

src/ec/packages/ec-component-button/button.story.js
```diff
@@ -28,15 +28,53 @@ uiIcons.forEach(icon => {
...
-  const dataWithIcon = data;
-  dataWithIcon.icon = {
-    type: 'ui',
-    path: defaultSprite,
-    size: 'xs',
+// Labels for the groups.
+const requiredGroupId = 'Mandatory elements';
+const optionalGroupId = 'Optional elements';
+
+// Preserve the adapted specs.
```
planctus: @Joosthe , why do you need this "cloning" of the input data? This is something we need only when we create buttons to hide/show something, because we need to retrieve back the original value from the specs, but here i don't see why we would need it. Apparently you are applying the complexity of other use cases where we needed to find solutions to specific issues, also to the standard use cases like this. You don't need to apply logic to this knobs, there is no such use case like the "label" will not be there, so why defining the knob like this? To me you should reuse everything from the data you have and only override what you need, because you want to set a knob, for instance, by redefining all the properties in the object instead you then need to set again things that are already in place (icon, for instance).
Doing like this, you would receive the data and you would then return the same data after applying the changes you want, like in the adapter, here you don't know anymore what was in the data that came from the specs, instead.

src/ec/packages/ec-component-button/button.story.js
planctus: @Joosthe, sorry, but nothing has changed regarding this from what i see. Is there anything unclear?
Not sure it was already the case before, also, but if you look at the button without the icon you will see that the markup of the icon is in place and the space is taken, just that there is no icon with that "name".

PR #348, opened 2020-02-27
feat(footers): Aligning with 2.24.0 FRONT-835

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-footer-core/ecl-footer-core.html.twig
```diff
@@ -40,49 +40,56 @@
...
+    {% if section.section_class_name is defined and section.section_class_name is not empty %}
+      {% set _section_class = _section_class ~ ' ' ~ section.section_class_name %}
+    {% endif %}
+    <section class="{{ _section_class }}">
+    {# content before #}
+    {% if section.content_before is defined and section.content_before is not empty %}
+      <div class="ecl-footer-core__content">{{ section.content_before|raw }}</div>
+    {% endif %}
+    {# title #}
+    {% if section.title is defined and section.title is not empty %}
```
Joosthe: Why are we declaren first local variable _sections where we afterwards are going to use section in the for loop. Doesn't it make more sense to decare this in the for loop aswell and make it {% for key, _section in _sections %} ?

src/ec/packages/ec-component-footer-core/footer-core.story.js
```diff
@@ -1,14 +1,113 @@
...
+    path: text(
+      'sections[0].title.link.path',
+      data[0].title.link.path,
+      requiredGroupId
+    ),
+    placeholder: text(
+      'optional elements',
+      'no optional elements defined in this story.',
+      optionalGroupId
+    ),
```
Joosthe: I would like a better label for the knobs. even if we make the label for developers this should make more sense. Also i don't see the knob  for " the description"  underneath the footer title.

src/ec/packages/ec-component-footer-harmonised/footer-harmonised.story.js
```diff
@@ -35,18 +31,399 @@ groups.forEach(g => {
...
+    dataG1[1][1].section_id = 2;
+    dataG1[2] = [{ section_id: 3 }, { section_id: 3 }];
+  }
+  // Site name & content owner details.
+  dataG1[0].title.link = {
+    label: text(
+      'sections[0].title.link.label',
+      dataG1[0].title.link.label,
+      requiredGroupId
+    ),
```
Joosthe: I think here also a better name for the sections and a beter order in the sections should help here. You can always look the find the knob by its value but still would it be nice if this was in order.

src/ec/packages/ec-component-footer-standardised/footer-standardised.story.js
```diff
@@ -1,29 +1,279 @@
-/* eslint-disable no-param-reassign */
+/* eslint-disable no-param-reassign, no-shadow */
 import { storiesOf } from '@storybook/html';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
 import withCode from '@ecl-twig/storybook-addon-code';
+import { withKnobs, button, text } from '@storybook/addon-knobs';
 import defaultSprite from '@ecl/ec-resources-icons/dist/sprites/icons.svg';
 
-import data from './demo/data';
+import sections from './demo/data';
```
Joosthe: why do we differ here from other components should we not have a standerd way of naming our data ?

src/ec/packages/ec-component-footer-core/ecl-footer-core.html.twig
planctus: Ok, yes, make sense.

src/ec/packages/ec-component-footer-core/footer-core.story.js
planctus: and do you have any better proposal? As we know this is sub-optimal anyhow, we need to define it just in order to see the tabs (mandatory/optional) when we have no optional element, that's why it literally says what we want to "communicate", that there is no optional element. Open to alternative solutions but unless we don't have a better one i think we need it.

src/ec/packages/ec-component-footer-harmonised/footer-harmonised.story.js
planctus: I appreciate the fact that you have in mind something better than this naming, but why don't you tell me? Saying "a better name would help" means that you have an idea on what it could be, generally speaking what i aim for is that we define the labels for knobs using the real property in the real "object" we send as an input data. In this case i see that i haven't been consistent in the three footers, but the example you put here is the "right" one, 
`sections[0].title.link.label`
is the real way to access that property in the input data, indeed while sometimes i wrongly used the name of the variable used in the story, i will change this.

src/ec/packages/ec-component-footer-standardised/footer-standardised.story.js
planctus: i'd say not at this level, @Joosthe , ok for consistency but it's normal that we create variables for them to mean something for us, this is the simplest use case, you only have one "data" object you get from the specs, but most of the times you have many, what about the consistency then, we name everything data1, data2, data3...? So, you are right in general in looking for consistency, but that doesn't mean we are going to have everything "aligned", stories are "individual" by definition, components are all different, being too strict is not going the help us.

planctus:
Thanks @Joosthe for the review, you can find in the last commits, separately, all the operations done after the review, local variables used everywhere, knob added in footer core, alignment of the knobs labels, basically... ;)

src/ec/packages/ec-component-footer-standardised/footer-standardised.story.js
Joosthe: I totally agree that we choose variable names that mean something to us. I don't mean strict the sense like data1, data2 , data3 this could be dataPrimary, dataSecondary, dataCall, ... . but i this that we should call these variables always data to begin with and your instance the default data in buttons should always and in data. For example:
`const primaryData = { ...dataPrimary };`. I don't see the added value to call data sections

src/ec/packages/ec-component-footer-harmonised/footer-harmonised.story.js
Joosthe: I don't think we need to change it completely but for instance we could call it `Section 0 link label` and if we have multiple links in the section we could call it `Section 0 link 1 label` and `Section 0 link 1 path`. To me this would look nicer but if you still want to go with sections[0].title.link.label its fine by me as well

PR #355, opened 2020-03-05
refactor(demo): Page Header Standardised, grouping knobs - FRONT-898

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


Joosthe:
If i use the extra-classes (separated by comma), the comma will be displayed in the html. The solution for this is or to separate by space of to use the array knob.

src/ec/packages/ec-component-page-header-standardised/page-header-standardised.story.js
```diff
@@ -9,17 +11,46 @@ import demoMetaTitleDescriptionContent from './demo/data--meta-title-description
...
+  }
+  if (data.description) {
+    data.description = text('description', data.description, optionalGroupId);
+  }
+
+  data.extra_classes = text(
+    'extra_classes (comma separated)',
+    '',
+    optionalGroupId
+  );
```
Joosthe: If i use this knob the html is update with a comma in between classes. The solution is to separate by space or to use the array knob

src/ec/packages/ec-component-page-header-standardised/page-header-standardised.story.js
```diff
@@ -9,17 +11,46 @@ import demoMetaTitleDescriptionContent from './demo/data--meta-title-description
...
+  data.extra_attributes = object(
+    'extra_attributes',
+    { name: '', value: '' },
+    optionalGroupId
+  );
```
Joosthe: This doesn't work in the when in check the html

src/ec/packages/ec-component-page-header-standardised/page-header-standardised.story.js
planctus: ah true, @Joosthe , but what is wrong here is just that label saying (comma separated), because it is not what we expect for extra classes, they have been always defined as a single string, so yes, "space" separated but that's not even worth to be mentioned. I'll change this everywhere ;)

src/ec/packages/ec-component-page-header-standardised/page-header-standardised.story.js
planctus: i'll check it

PR #357, opened 2020-03-06
refactor(demo): Checkbox, optional/mandatory knobs - FRONT-901

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


Joosthe:
just for checking but why are we displaying  on select knobs with only one item ? Isn't there a better way to display the value without making it changeable ?

planctus:
> just for checking but why are we displaying on select knobs with only one item ? Isn't there a better way to display the value without making it changeable ?

Eh, no. This is the solution i've found after trying to see whether there was any way in storybook to define a knob as a disabled input field.
The reason why there is no standard way of doing this in storybook is because it's a bit against the idea of the knob itself, to make it "disabled", indeed here what we do is to use the only "input field" that let us show an information without making it editable, i think is a smart solution.. :)

PR #358, opened 2020-03-06
refactor(demo): Radio, grouping knobs - FRONT-902

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-radio/radio.story.js
+++ b/src/ec/packages/ec-component-radio/radio.story.js
@@ -18,26 +18,48 @@ import notes from './README.md';
 // Labels for the groups.
 const requiredGroupId = 'Mandatory elements';
 const optionalGroupId = 'Optional elements';
+const statesGroupId = 'States';
 
 const prepareRadio = data => {
+  data.invalid = boolean('invalid', false, statesGroupId);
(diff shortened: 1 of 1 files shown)


src/ec/packages/ec-component-radio/radio.story.js
```diff
@@ -8,46 +9,140 @@ import dataBinary from './demo/data--binary';
...
+      optionalGroupId
+    );
+    data.optional_text = text(
+      'optional_text',
+      data.optional_text,
+      requiredGroupId
+    );
+  }
+  data.helper_text = text('helper_text', data.helper_text, optionalGroupId);
+
```
Joosthe: The only thing about this component, is when you pass in a special character and you find, the unicode in the html. for all the rest the component  looks perfect

planctus: you're right, but this looks like a more general issue we have when we pass the json to the notes tab, it's not only about this knob. Trying to find a solution but no luck yet, think we can live with this for now if we don't find a straightforward solution, though.

planctus: I've found actually something that works, but as i mentioned this is not only about this single knob, so it's something we have to fix in the notes plugin.

PR #362, opened 2020-03-12
fix(button): Fixing extra_attributes knobs, adding states group - FRONT-947

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-button/button.story.js
```diff
@@ -37,30 +31,59 @@ uiIcons.forEach(icon => {
...
+      '',
+      optionalGroupId
+    );
+    attribute.name = attribute1Name;
+    if (attribute1Value !== '') {
+      attribute.value = attribute1Value;
+    }
+    data.extra_attributes.push(attribute);
+    // Second attribute.
+    if (attribute2Name !== '') {
```
papegaill: Is it really necessary to be able to set a second attribute? 
My feeling is that we are having more and more code. As extra_classes and extra_attributes is part of every-components, we can maybe have these knobs more globally?

planctus: It's true, we need code to implement functionalities.. :)
I though about having these knobs, which are always the same from a utility sort of, but i don't know, that could be tricky, actually, and most likely misbehaving, we need to define these knobs in the stories, but we can give it a try. 
The reason why i don't feel disturbed by the "code" needed to implement these "optional knobs"for extra_classes and extra_attribute is that it's always the same and if we decide to go for a "utility" for instance, this is a standard change to be applied to all the components with no particular effort.
I'll give it a try and see what consequencies this might have, for the "second" attribute yes, i think is important to show the fact that this parameter accepts an array and you can set multiple extra_attributes.

papegaill: ok @planctus are you creating a ticket to investigate utilities,
if yes I can merge this and we will see if this can be improved globally?

planctus: ah well, no i did it here and it seems to work.. :)  
if we like the solution we merge it and i adapt all the other pull requests opened, i think we will add also other functions to this node module.. ;)
To make it work you need to  run yarn

utils/story-utils/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+export const getIconKnobs = (data, name, size, color, transform) => {
+  const sizes = {
+    XS: 'xs',
+    S: 's',
+    M: 'm',
+    L: 'l',
+    XL: 'xl',
+    '2XL': '2xl',
```
papegaill: a reason why this one is quoted?

planctus: try without :)

PR #373, opened 2020-03-17
refactor(demo): hero-banner optional/mandatory knobs - FRONT-912

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
+  getExtraKnobs(data);
+  if (data.link.link.label) {
+    data.link.link.label = text(
+      'Link label',
+      data.link.link.label,
+      buttonLabels.required
+    );
+  }
+
+  if (data.title) {
```
planctus: @Joosthe, you duplicated this knob here. But anyway, i don't understand the reason why of the check you do for the title and the description, if am not wrong they are in all the variants no..? So you can remove the ifs for them.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
+  if (data.description) {
+    data.description = text(
+      'Description',
+      data.description,
+      buttonLabels.required
+    );
+  }
+
+  if (data.centered) {
+    data.centered = boolean('Centered', data.centered, buttonLabels.optianal);
```
planctus: I believe this should be buttonLabels.optional in your intention :)
But at the same time, although is not entirely proper, i would place this a "state".
At the moment, because of this typo, you have the "other" tab in the knobs, this will disappear once you apply the right label.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
 import notes from './README.md';
 
 const PrepareBanner = data => {
+  getExtraKnobs(data);
```
planctus: So, it turns out that we have, actually, control on the order of the knobs, it's just the way you place them in the code.
That's why you saw this getExtraKnobs call always at the end of the processing function, because these are normally the last ones we want to show ;)

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
+      buttonLabels.required
+    );
+  }
+
+  if (data.title) {
+    data.title = text('Title', data.title, buttonLabels.required);
+  }
+
+  if (data.title) {
+    data.title = text('Title', data.title, buttonLabels.required);
```
planctus: You see also that the effort is to use labels for the knobs that are corresponding to the right piece of information in the input data, this property is title, not Title and so its label should be ;)

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
+      buttonLabels.required
+    );
+  }
+
+  if (data.centered) {
+    data.centered = boolean('Centered', data.centered, buttonLabels.optianal);
+  }
+
+  if (data.image) {
+    data.image = text('Image', data.image, buttonLabels.optianal);
```
planctus: Well, also here you have the same typo as before and the label should be image, not Image

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
+  }
   data.link.icon.path = defaultSprite; // eslint-disable-line no-param-reassign
   return data;
```
planctus: Would you insert a blank line (with no spaces) before the return, please?

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
 import notes from './README.md';
 
 const PrepareBanner = data => {
+  getExtraKnobs(data);
+  if (data.link.link.label) {
```
planctus: This order is valid also for the tabs, ideally i would imagine to see "states" | "mandatory elements" | "optional elements", in this order. There might be situation where this could be tricky, but that's the goal.. ;)
This means that if the first knob you define is withing a group, this group will also be the first tab on the panel.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
planctus: Would you add also the other knobs for the link? the path, the label..?

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
...
-          image: text('Image', dataImage.image),
-        })
-      ),
+    () => {
+      const bannerImage = PrepareBanner(dataImage);
+
+      return heroBanner(bannerImage);
+    },
     {
       notes: { markdown: notes, json: PrepareBanner(dataImage) },
```
planctus: MMhh...i have the impression you did not follow the pattern here, you can use here dataImage for how it is, here in the notes, basically

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -15,105 +16,97 @@ import heroBanner from './ecl-hero-banner.html.twig';
 import notes from './README.md';
 
 const PrepareBanner = data => {
```
planctus: camelCase, @Joosthe , means first world starts with a lowercase...

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
planctus: This is still in place:
` if (data.title) `
` if (data.description) `
The question about these was: why? There is no need to check for these properties since they are always set.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
planctus: Ok, you fixed the typo but the "centered" parameter has not been moved to the "states" group.
The reason why i think is the right thing to do is that this option should be available for all the variants, but the "align-left" one (it's also like this in ECL). 
So, my recommendation would be to set the "centered" parameter for all the variants in the "states" group excluding the "align-left" one.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
planctus: still valid.

planctus: Still valid, the label is there (with a typo, though: link.Link.label) but the path is not, the icon path is not. We should define a knob for everything that can be set on the component, ideally, it has to be possible to customise the component in the demo with all the different possibilities and then being able to just copy the include code from the notes tab to reproduce the look and feel of the component for how you see it in storybook, that's the goal :)

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
planctus: Still valid.

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -14,107 +15,94 @@ import dataLeft from './demo/data--align-left';
...
+      buttonLabels.required
+    );
+  }
+
+  if (data.centered) {
+    data.centered = boolean('centered', data.centered, buttonLabels.optional);
+  }
+
+  if (data.image) {
+    data.image = text('image', data.image, buttonLabels.optional);
```
planctus: I don't think this should be "optional", well it is "optional" in the general definition of the component, but since the demo we have for the banner is showing all the different variants including the ones without an image (primary and default) in the "image" variant of the component the image is actually mandatory.
So, it should not be possible to see the "image" variant without an image, the result is wrong, because it should actually fallback to one of the aforementioned ones (primary or default).
Furthermore, i don't see any knob for the fundamental parameter "type". This is an optional parameter (since we have a default value for that) and we should show it in a select element with the current type option only. ;)

src/ec/packages/ec-component-hero-banner/hero-banner.story.js
```diff
@@ -14,107 +15,94 @@ import dataLeft from './demo/data--align-left';
...
+  }
   data.link.icon.path = defaultSprite; // eslint-disable-line no-param-reassign
```
planctus: Ah, so this should turn into a knob select with the only option being the current path for the icons file.
But you have the exclusion for this eslint check at the top of the file, you don't really need to exclude the check in this line in particular, @Joosthe

Joosthe:
@planctus Looks really much improved and thanks for all the work, the only thing i could find on the component is that the icon is missing now in the preview of the component

planctus:
No, @Joosthe , the icon is there:
https://5e73a38ee6b9de0ee0b17d46--ecl-twig.netlify.com/ec/?path=/story/components-banners-hero-banner--image
I think you just need to update the plugin for the story-utils, sorry not to point this out in advance, but as usual, when we have changes in one of our plugins we need to pay attention to really update it on our local environments.
So, you can remove the node-modules folder and run yarn again or you can copy the index.js file from the working tree (utils/story-utils/src/index.js) to the package in the node_modules folder (node-modules/@ecl-twig/story-utils/src/index.js).
I hope this is the last time i have to write this, but it's true that might maybe find a solution for this beacuse i see that everytime is an issue.

PR #375, opened 2020-03-17
refactor(demo): Messages, optional/mandatory knobs - FRONT-916

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+    iconsList,
+    data.icon.name,
+    buttonLabels.required
+  );
+  if (name !== null) {
+    getIconKnobs(data, name, 'notifications', 'l');
+  }
+
+  if (data.title) {
+    data.title = text('Title', data.title, buttonLabels.required);
```
planctus: the label should be title, no?

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+  }
+
+  if (data.description) {
+    data.description = text(
+      'Description',
```
planctus: Also here, the parameter is description, then we spell it as such also in the label ;)

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
 import notes from './README.md';
 
+const iconsList = {};
+iconsList.none = null;
+
+notifIcons.forEach(icon => {
+  iconsList[icon] = icon;
+});
+
+const PrepareMessage = data => {
```
planctus: camelCase, @papegaill ;)

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+  if (data.description) {
+    data.description = text(
+      'Description',
+      data.description,
+      buttonLabels.required
+    );
+  }
+  getExtraKnobs(data);
+
+  return data;
```
planctus: We should have also a knob for the variant, of type select, with the only option being the current one, since we have a story for each of them.

Also the thing you are now doing in formatIcon should actually be part of the main prepare function where you set them as knobs, still select with that only option, as mandatory, without these you don't render the component correctly otherwise.
This will also let you use in the json parameter of the notes the dataInfo, dataWhatever directly with no processing.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
 
+const iconsList = {};
+iconsList.none = null;
+
+notifIcons.forEach(icon => {
+  iconsList[icon] = icon;
+});
+
+const PrepareMessage = data => {
+  const name = select(
```
planctus: Mmhh, i think this should not be editable, we have all the stories for all the variants, therefore this should be a select with the only option being the current icon.name. Also because this at the moment doesn't make much sense, you can basically apply the wrong icon to a certain variant.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+    'icon.name',
+    iconsList,
+    data.icon.name,
+    buttonLabels.required
+  );
+  if (name !== null) {
```
planctus: Well, there is not going to be any null case for the name, so...

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+    getIconKnobs(data, name, 'notifications', 'l');
+  }
+
+  if (data.title) {
```
planctus: But why the if..? Do we have the possibility of this title not being passed in the story? No, it can only be an empty string and we don't need to check anything here.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+    data.title = text('Title', data.title, buttonLabels.required);
+  }
+
+  if (data.description) {
```
planctus: same here, it's mandatory, we set a knob for it, no need to check for it to be present

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,6 +19,40 @@ import dataWarning from './demo/data--warning';
...
+  const name = select(
+    'icon.name',
+    iconsList,
+    data.icon.name,
+    buttonLabels.required
+  );
+  if (name !== null) {
+    getIconKnobs(data, name, 'notifications', 'l');
+  }
+
```
planctus: I don't see the color acting in a anyway and i don't think we should give the chance to edit the transform either, here it's one icon per variant, no "options" whatsoever, so pass also the other parameters, i'd sy, to the function giving you the icon knobs.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -25,67 +63,15 @@ storiesOf('Components/Messages', module)
...
+    notes: { markdown: notes, json: formatIcon(dataInfo) },
+  })
+  .add('Success', () => message(PrepareMessage(dataSuccess)), {
+    notes: { markdown: notes, json: formatIcon(dataSuccess) },
+  })
+  .add('Error', () => message(PrepareMessage(dataError)), {
+    notes: { markdown: notes, json: formatIcon(dataError) },
+  })
+  .add('Warning', () => message(PrepareMessage(dataWarning)), {
+    notes: { markdown: notes, json: formatIcon(dataWarning) },
```
planctus: very cute.. ;)

papegaill:
@planctus Many thank for the review, PR updated accordingly ;)

planctus:
Ooook, much much better but:
at the moment we lost the close icon, i guess it's because we also miss a knob related to it where we would override the default "icons.svg" path we normally get from the adapters.

The other small thing is that i would expect title and description knobs to appear before the ones for the icon.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,77 +17,52 @@ import dataWarning from './demo/data--warning';
...
 
-const formatIcon = data => {
-  data.icon.path = defaultSprite;
-  data.close.icon.path = defaultSprite;
+const prepareMessage = data => {
+  const variantList = {};
+  variantList[data.variant] = data.variant;
+  data.variant = select(
+    'variant',
+    variantList,
```
planctus: i don't neessarily understand this, @papegaill ..wouldn't be enough to create the select like this?
`data.variant = select('variant', [data.variant], data.variant); `

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,77 +17,52 @@ import dataWarning from './demo/data--warning';
...
+    data.variant,
+    buttonLabels.required
+  );
+
+  const iconsList = {};
+  iconsList[data.icon.name] = data.icon.name;
```
planctus: same as before.. ;)

papegaill:
@planctus, many thx for suggestion ! PR updated.

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,77 +18,55 @@ import dataWarning from './demo/data--warning';
...
+    'icon.name',
+    [data.icon.name],
+    data.icon.name,
+    buttonLabels.required
+  );
+  getIconKnobs(data, name, 'notifications', 'l', 'primary', 'none');
+
+  data.title = text('title', data.title, buttonLabels.required);
+
+  data.description = text(
```
planctus: Why don't we move these knobs on top? title and description i mean?

src/ec/packages/ec-component-message/message.story.js
```diff
@@ -15,77 +18,55 @@ import dataWarning from './demo/data--warning';
...
+    'description',
+    data.description,
+    buttonLabels.required
+  );
+
+  data.close.icon.path = select(
```
planctus: I think we need also something else for the close button, or maybe we should set it in the template, i'm referring to the icon.size for the close button...

src/ec/packages/ec-component-message/message.story.js
papegaill: @planctus I was following same order as https://github.com/ec-europa/ecl-twig/blob/4b67ca05606c3444040e1be7bc7290af40ee64c9/src/ec/packages/ec-component-message/ecl-message.html.twig#L5

But yes maybe, it better to have them on top?

src/ec/packages/ec-component-message/message.story.js
papegaill: Sure, updated, I think it's best to have it in the template as you suggested.

src/ec/packages/ec-component-message/message.story.js
planctus: ah...i see the criteria, but yes i think we want them on top regardless of how we wrote this "example".. :)

src/ec/packages/ec-component-message/message.story.js
planctus: ook

src/ec/packages/ec-component-message/message.story.js
papegaill: ok make sense, PR updated ;)

PR #378, opened 2020-03-18
refactor(demo): Tags, optional/mandatory knobs - FRONT-922

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


planctus:
@papegaill , i noticed that we had a linter issue in the build. Now this might probably mean that something is wrong with your pre-commit hook, since it should have been blocked by husky and it was not. Would you check on your side the situation? You should have a pre-commit file in the .git/hooks folder, inside you should have code from husky and at the end the grumphp one. When we commit both checks should be in place, but it's easy to break this configuration, unfortunately, also in my case yesterday i pushed a console.log() statement and the build was even green which makes me think that some updates might have broken our configuration...

papegaill:
@planctus yes I have these husky files in my git/hook folder.
And grumpphp sniffing my code before every commit, 
```
GrumPHP detected a pre-commit command.
GrumPHP is sniffing your code!
```
And everything was "all good" on my side.

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,45 @@
...
 import dataLink from './demo/data--link';
 import dataButton from './demo/data--button';
 import dataRemovable from './demo/data--removable';
 import tag from './ecl-tag.html.twig';
 import notes from './README.md';
 
+// Preserve the adapted specs.
+const prepareTag = data => {
+  data.default_icon_path = defaultSprite;
+  data.tag.label = text('label', data.tag.label, buttonLabels.required);
```
planctus: but if it's
`data.tag.label` the label is expect is:
`tag.label`
You have t think that this is the property in the input data

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,45 @@
...
+  if (data.tag.path) {
+    data.tag.path = text('Url', data.tag.path, buttonLabels.required);
```
planctus: And then the label for this would be
`tag.path`

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,45 @@
...
+  }
+  if (data.tag.aria_label) {
+    data.tag.aria_label = text(
+      'Aria label',
```
planctus: This one's label:
`tag.aria_label`
But the thing is that this is always an optional parameter, not only when it is set in the input data ;)

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,45 @@
...
+      data.tag.aria_label,
+      buttonLabels.required
+    );
+  }
+  getExtraKnobs(data);
+
+  return data;
```
planctus:
```
{#
  Parameters:
   - "tag" (associative array) (default: predefined structure): format:
      {
        type: 'link' (can be 'link', 'button', 'removable'),
        path: '' (tag url if needed),
        label: '' (tag text as string),
        aria_label: '' (optional) (aria label for removable tag as a string)
      }
    - "default_icon_path" (string) Default path for icon tag (need to render Icon component)
    - "extra_classes" (string) (default: '')
    - "extra_attributes" (array) (default: []): format: [
        {
          "name" (string) (default: ''),
          "value" (optional) (string)
        ...
      ]
#}
```

The idea would be that through the knobs you would be able to rebuild the entire input data, actually the see it interacting with it when it makes sense to do it, if you want to edit things for then copying the include code in the notes tab.
So here for the tag we would also have:
```
type as a select with the only option being the current variant
default_icon_path being again a select with the only option being the current value
```

papegaill:
@planctus Thank you, PR updated accordingly.

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,62 @@
...
 import dataLink from './demo/data--link';
 import dataButton from './demo/data--button';
 import dataRemovable from './demo/data--removable';
 import tag from './ecl-tag.html.twig';
 import notes from './README.md';
 
+// Preserve the adapted specs.
+const prepareTag = data => {
+  const typeList = {};
+  typeList[data.tag.type] = data.tag.type;
```
planctus: It all works fine, but i still don't understand this approach here for building a select with a single element. You don't need any processing, you just do this:
```
data.type = select(
    'tag.type',
    [data.tag.type],
    data.tag.type,
    buttonLabels.required
  );
```
Look that right now you are showing the right information, but it refers to the wrong piece of data, data.variant doesn't exist, is actually data.type, and doing this introduces a duplicated parameter in the notes tab.

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -1,63 +1,62 @@
...
+  }
+  if (data.tag.aria_label) {
+    data.tag.aria_label = text(
+      'tag.aria_label',
+      data.tag.aria_label,
+      buttonLabels.required
+    );
+  }
+  if (data.default_icon_path) {
+    const iconPathList = {};
```
planctus: This is a pattern, you did the same  everywhere and you can simplify this, in this case it would be:
```
data.default_icon_path = select(
      'default_icon_path',
      [defaultSprite],
      defaultSprite,
      buttonLabels.required
    );
```

src/ec/packages/ec-component-tag/tag.story.js
papegaill: Thx !

PR #381, opened 2020-03-18
refactor(demo): Accordion(2) knobs and labels added - FRONT-910

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -3,32 +3,39 @@ import { storiesOf } from '@storybook/html';
...
 import withCode from '@ecl-twig/storybook-addon-code';
-
+import { buttonLabels, getExtraKnobs } from '@ecl-twig/story-utils';
 import defaultSprite from '@ecl/ec-resources-icons/dist/sprites/icons.svg';
 import demoData from './demo/data';
 
 import accordion2 from './ecl-accordion2.html.twig';
 import notes from './README.md';
 
+const PrepareAccordion2 = data => {
```
planctus: camelCase, @Joosthe

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -3,32 +3,39 @@ import { storiesOf } from '@storybook/html';
...
+  data.items.forEach((item, index) => {
+    const { content, toggle } = item;
+
+    item.toggle.label = text(
+      `toggle.label ${index}`,
```
planctus: You saw what is the result of this?
`toggle.label 0`
do you recognise it as anything that would work in code? Clearly is not, but more than that you would still miss the items array where these properties have to be set, the label is supposed to be the "code to access it", not sure i'm able to convince you about this, but this should be:
`items[${index}].toggle.label`

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -3,32 +3,39 @@ import { storiesOf } from '@storybook/html';
...
+      toggle.label,
+      buttonLabels.required
+    );
+
+    item.content = text(`content ${index}`, content, buttonLabels.required);
+    item.toggle.icon.path = defaultSprite;
```
planctus: this should be a knob, everything should be a knob, not sure what can still be unclear about what we are doing in this "demo" track but i'll try again: we are trying to expose EVERY thing you can do with a component, every single property you can or you have to set, giving all the possibilities to customise the demo of the component and then reporting the correct code for the include in the notes tab.
if you have to set a a path to an icon to make an icon work i want that to be a knob, (a select knob with only that option) and this is valid for all the rest, there are now plenty of examples of this being done in multiple components.

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -3,32 +3,39 @@ import { storiesOf } from '@storybook/html';
...
-        const { id, content, toggle } = item;
-
-        item.toggle.label = text(`Label ${id}`, toggle.label);
-        item.content = text(`Content ${id}`, content);
-
-        item.toggle.icon.path = defaultSprite;
-      });
-
-      return accordion2(demoData);
+      const accordion2DemoData = PrepareAccordion2(demoData);
```
planctus: Basing on the work @papegaill made, apparently in the easiest scenario, which this look like, we can avoid the separate call for the prepare function for then returning the component. You can see an example of what i'm saying in the messages component.
Not a needed change in this case, but since it looks like a simplified solution, the one that @papegaill applied, maybe we can opt for that when we don't need to further customise the data in the story we add.

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -1,36 +1,49 @@
...
 import demoData from './demo/data';
 
 import accordion2 from './ecl-accordion2.html.twig';
 import notes from './README.md';
 
+const prepareAccordion2 = data => {
+  data.items.forEach((item, index) => {
+    const { content, toggle, id } = item;
+
+    item.id = select(`item ${index} id`, [id], id, buttonLabels.required);
```
planctus: - These labels are supposed to be the way in code you would retrieve a certain prop:

`item ${index} id` produces `item 0 id` which is not the way you would write the code to access the property, this is instead:
`items[0].id`
because your object is named items and it's an array, item is just a conventional name you used in the forEach, and you get this format by writing:
`items[${index}].id`

- The same is valid for all the others label you generated but:

- iconShape for sure is not something we pass to the template as such


- i don't see the icon_path knob and i don't see the "level" knob either, which were the changes requested in the previous review, @Joosthe so i'll leave also the previous "conversation" open.

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -1,36 +1,49 @@
 /* eslint-disable no-param-reassign */
 import { storiesOf } from '@storybook/html';
-import { withKnobs, text } from '@storybook/addon-knobs';
+import { withKnobs, text, select } from '@storybook/addon-knobs';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
 import withCode from '@ecl-twig/storybook-addon-code';
-
+import { buttonLabels, getExtraKnobs } from '@ecl-twig/story-utils';
```
planctus: We try to separate the import of objects of functions from the one about specs and resources, just by an empty line.

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -1,36 +1,51 @@
...
 import accordion2 from './ecl-accordion2.html.twig';
 import notes from './README.md';
 
+const prepareAccordion2 = data => {
+  data.items.forEach((item, index) => {
+    const { id, level, toggle, content } = item;
+
+    item.id = select(`item[${index}].id`, [id], id, buttonLabels.required);
+    item.level = select(
+      `item[${index}].level`,
```
planctus: @Joosthe..can you take a look at the demo data? Item doesn't exist, i wrote you this already.
This is:
`items[${index}].level`

and the same applies to the other knobs. 

**But this should be editable, the level is the heading that will be generated, this should be a select, yes, but with six values (1, 2, 3, 4, 5, 6) with the default set to 3 (as it is in the template), at least basing on the template we have.**

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -1,36 +1,51 @@
...
+      defaultSprite,
+      buttonLabels.required
+    );
+    item.content = text(
+      `item[${index}].content`,
+      content,
+      buttonLabels.required
+    );
+  });
+  getExtraKnobs(data);
```
planctus: can you add an empty line before the return, please?

src/ec/packages/ec-component-accordion2/accordion2.story.js
```diff
@@ -1,36 +1,51 @@
...
+    const { id, level, toggle, content } = item;
+
+    item.id = select(`item[${index}].id`, [id], id, buttonLabels.required);
+    item.level = select(
+      `item[${index}].level`,
+      [level],
+      level,
+      buttonLabels.required
+    );
+    item.toggle.label = text(
```
planctus: So, nice that yo uadded the missing knobs, but i think we can improve the order, logically i would expect the "items[i].content" to come earlier than the "items[i].toggle.icon.path", no?

src/ec/packages/ec-component-accordion2/accordion2.story.js
planctus: can you please do just the same that is done in the deprecated accordion, regarding the organisation of this import statements? It's a minor thing, eh, but maybe exactly because of this we shouldn't let it go.

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: Sure i changed the order i just put it like the order from the notes.

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: Sure, but i the ordered them the same the order from the notes

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: oke i've made an array for this.

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: oke i have re-ordered them in this import statement

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: oké is done

src/ec/packages/ec-component-accordion2/accordion2.story.js
papegaill: @Joosthe please check this remarks https://github.com/ec-europa/ecl-twig/pull/381#discussion_r396542121

src/ec/packages/ec-component-accordion2/accordion2.story.js
papegaill: @Joosthe please update this

Joosthe: @papegaill  Ok

src/ec/packages/ec-component-accordion2/accordion2.story.js
Joosthe: I did this exactly like the deprecated accordion or what do you mean by this ?

papegaill: ok indeed seems good to me, https://github.com/ec-europa/ecl-twig/pull/381#discussion_r396542121 was not clear to me.

planctus:
i see this:
image

To me the order should be, instead:
level,
id, 
content,
toggle.label
toggle.icon.path

Joosthe:
Kobs are re-orderd and ticket can be closed

PR #382, opened 2020-03-19
refactor(demo): Inpage navigation, optional/mandatory knobs - FRONT-927

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


planctus:
again an issue with the linter in the story file, there has to be something wrong..

src/ec/packages/ec-component-inpage-navigation/inpage-navigation.story.js
```diff
@@ -72,26 +74,37 @@ const btnIdHandler = () => {
...
 
+// Preserve the adapted specs.
+const prepareNavigation = data => {
+  // Buttons for the demo.
+  button(btnLeftLabel, btnLeftHandler, buttonLabels.cases);
+  button(btnMainLabel, btnMainHandler, buttonLabels.cases);
+  button(btnIdLabel, btnIdHandler, buttonLabels.cases);
+  button(btnIdRemoveLabel, btnIdRemoveHandler, buttonLabels.cases);
+
+  data.title = text('Title', data.title, buttonLabels.required);
```
planctus: the label should be:
`title`

src/ec/packages/ec-component-inpage-navigation/inpage-navigation.story.js
```diff
@@ -72,26 +74,37 @@ const btnIdHandler = () => {
...
+  data.links = object('Links', data.links, buttonLabels.required);
```
planctus: here it should be
`links`
but no, ok, this was my idea but like these, is wrong, we pollute the data to build the demo, but this should be:
```
- "links" (array) (default: predefined structure): format: [
      {
        href: (string) (default: ''),
        label: (string) (default: '')
      },
      ...
      ]
```
So what we should do is to create two text knobs for each link that would be like:

`links[${i}].href` 
and
`links[${i}].label`

but you have to see what happens than if you use the buttons, to add or remove items from the inpage, i guess that the knobs would not be updated, but still it's acceptable.

src/ec/packages/ec-component-inpage-navigation/inpage-navigation.story.js
```diff
@@ -72,26 +74,37 @@ const btnIdHandler = () => {
...
-      button(btnMainLabel, btnMainHandler);
-      button(btnIdLabel, btnIdHandler);
-      button(btnIdRemoveLabel, btnIdRemoveHandler);
-
       let pageFillerHtml = '';
       demoData.links.forEach(content => {
         pageFillerHtml += content.item;
       });
 
       const fullDemoData = { ...demoData, icon_path: iconPath };
```
planctus: Ahhh, this has to be a knob:
`icon_path`
if you don't set this you don't implement the component correctly, so it's also a mandatory one, a select with a single option...

src/ec/packages/ec-component-inpage-navigation/inpage-navigation.story.js
papegaill: done @planctus, yes knobs are not updated if we play with the buttons to add or remove items.
This can be confusing in the end, I think..  maybe we should remove these buttons, what do you think?

papegaill:
> When i try to inject a new heading i get a heading with title "New heading 7s8uy". Why do you set the hash in the tilte or is 7s8uy is something else ?

I was already the case @planctus?

papegaill:
> When i try to inject a new heading i get a heading with title "New heading 7s8uy". Why do you set the hash in the tilte or is 7s8uy is something else ?

@Joosthe As new heading is inserted in a random order, this hash is used to generate a unic random id for the new content added in the navigation/content.

planctus:
MMhh...now that i see the whole i understand more your initial concerns, maybe, most likely we are not adding big value to this component demo by doing this, anyway, but we could maintain the knobs for how they are now, instead we absolutely have to prevent the madness we do to build the demo of this component to appear as the include code for the component in the notes tab.
For doing this it would be enough to just not pass the json parameter in the notes but at the same time at the moment the changes in the knobs are not reflected there, including extra_classes and extra_attributes and they are not even reflected in the html and if we remove the json parameter we would never see them, so, three options:

1.  we do some trick to clean up the code we pass to the notes tab (it can be done in the notes plugin itself) and we ensure the notes tab is only showing the links in the inpage, plus extra-classes and extra_attributes.

2. we pass in the json only the "pure links" of the inpage, plus extra_attributes and extra_classes, reflecting the real state of the demo.

3.  We pass a static object, plus extra_attributes and extra_classes retrieved from the knobs

what do you prefer? :)

Ah, there is also the "giving up" option where we just improve a bit the example in the readme and we don't set any knob at all but i didn't want to consider it for now :)


UPDATE:
There is something else wrong, i discovered that when you play with the knobs, like you add an extra_classes you don't see anything reflected in html or in the notes tab but at the same time you reset anything you might have been doing with the "use cases" (the knobs are resetting it) and furthermore if you then use any of the button in the use_cases you get unpredictable results, like multiple executions of the expected operation.
In other words these buttons should act on the data and not on the html, in order for any of the solutions before to work and i'm not sure we want to go there, i think we don't, still in this case the value of the "include code" is very limited, indeed.

At the end of the story i think that the "giving up" option could be actually considered. :(

papegaill:
@planctus As discussed, I removed the added knobs and reviewed the default notes.

PR #388, opened 2020-03-25
refactor(knobs): table optional/mandatory knobs - FRONT-924

Joosthe:
# PR description

There is a problem with the notes panel, basically for the way we set in the data the extra_attributes and extra-classes for the rows, when we go rendering that in the panel we lose these information.

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-table/table.story.js
+++ b/src/ec/packages/ec-component-table/table.story.js
@@ -1,37 +1,65 @@
+/* eslint-disable no-param-reassign */
 import { storiesOf } from '@storybook/html';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
+import { getExtraKnobs, buttonLabels } from '@ecl-twig/story-utils';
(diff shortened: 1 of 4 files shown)


src/ec/packages/ec-component-table/table.story.js
```diff
@@ -1,55 +1,96 @@
...
-      notes: { markdown: notes, json: dataWithRowExtraAttributes },
-    }
-  )
-  .add(
-    'Zebra',
-    () => {
-      dataDefault.zebra = true;
-      return table(dataDefault);
-    },
+    () => table(prepareTable(dataDefault, true)),
```
papegaill, line 86: why is zebra set to true for default table?

planctus: but this parameter is not about the zebra, why did you assume so? :)

papegaill: indeed, sorry for confusion @planctus, my bad.

PR #390, opened 2020-03-25
refactor(knobs): Datepicker - Mandatory/optional knobs -  FRONT-905

Joosthe:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-datepicker/datepicker.story.js
+++ b/src/ec/packages/ec-component-datepicker/datepicker.story.js
@@ -1,13 +1,55 @@
 /* eslint-disable no-param-reassign */
 import { storiesOf } from '@storybook/html';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
-import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
 import withCode from '@ecl-twig/storybook-addon-code';
+import { getExtraKnobs, buttonLabels } from '@ecl-twig/story-utils';
 
 import defaultSprite from '@ecl/ec-resources-icons/dist/sprites/icons.svg';
-import data from './demo/data';
+import dataDefault from './demo/data';
 import datepicker from './ecl-datepicker.html.twig';
 import notes from './README.md';
+
+const prepareDatePicker = data => {
+  data.label = text('label', data.label, buttonLabels.required);
+  data.icons_path = select(
+    'icons_path',
+    [defaultSprite],
+    defaultSprite,
+    buttonLabels.required
+  );
+  data.helper_text = text(
+    'helper_text',
+    data.helper_text,
+    buttonLabels.required
+  );
+  data.invalid = boolean('invalid', false, buttonLabels.cases);
+  data.invalid_text = text(
+    'invalid_text',
+    data.invalid_text,
+    buttonLabels.optional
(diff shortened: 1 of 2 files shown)


planctus:
@Joosthe this ticket:
https://webgate.ec.europa.eu/CITnet/jira/browse/FRONT-905
was flagged and with a comment saying that there is no spec page in ECL:
https://ec.europa.eu/component-library/ec/components/forms/datepicker/usage/

i see it has not been added so i wonder what you based this work on.

Joosthe:
Since the knobs were already there i set the knobs i based the cases on the variants. the optional elements are the elements that are visible if one of the cases is true

planctus:
Yep, @Joosthe , i thought i edited the comment replacing the "i see it has not been added so i wonder what you based this work on." with " i see that what you did makes sense, i will unflag the ticket" right after taking a look at the code, but i didn't publish it sorry, i removed the flag on the ticket instead, that i did it for real.. :)

src/ec/packages/ec-component-datepicker/datepicker.story.js
```diff
@@ -1,35 +1,66 @@
 /* eslint-disable no-param-reassign */
 import { storiesOf } from '@storybook/html';
 import { withNotes } from '@ecl-twig/storybook-addon-notes';
-import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
 import withCode from '@ecl-twig/storybook-addon-code';
+import { getExtraKnobs, buttonLabels } from '@ecl-twig/story-utils';
```
planctus: You might not have noticed this, but we have from the story-utils package also a function returning the form knobs which is what you might want to use here instead of redefining all the knobs in this story. 
`getFormKnobs()`

Furthermore, you set these boolean as "use cases" while in all the other form elements these are in the group "states", you will fix this by just using the right function from the utilities, by the way.

src/ec/packages/ec-component-datepicker/datepicker.story.js
```diff
@@ -1,35 +1,66 @@
...
 
 import defaultSprite from '@ecl/ec-resources-icons/dist/sprites/icons.svg';
-import data from './demo/data';
+import dataDefault from './demo/data';
 import datepicker from './ecl-datepicker.html.twig';
 import notes from './README.md';
 
+const prepareDatePicker = data => {
+  data.label = text('label', data.label, buttonLabels.required);
```
planctus: The order of the group knobs should be: States | Mandatory | Optional

src/ec/packages/ec-component-datepicker/datepicker.story.js
```diff
@@ -1,35 +1,66 @@
...
+    'required_text',
+    data.required_text,
+    buttonLabels.optional
+  );
+  data.optional_text = text(
+    'optional_text',
+    data.optional_text,
+    buttonLabels.optional
+  );
+  data.placeholder = text(
```
planctus: i think for now you can move this into the optional parameters, as we know there is no documentation page to base this on, but yeah, it looks more like an "option" than a requirement

planctus: You haven't changed this, but since there is no official docs about what is required and what is not for this component we can leave this as required.

PR #395, opened 2020-03-30
fix(pre-commit): Different approach for combining husky and grump-php - FRONT-960

planctus:
# PR description

## QA Checklist

So, the logic here is:
We now have a way to combine grumphp and husky, we first execute grump-php and basing on its output we run or not the husky pre-commit script.
This way we prevent the commit from happening when there is an issue in one of our templates, the whole library is checked and not only the "staged" files which would normally happen using the pre-commit hook provided by grump-php, but the total number of files to be checked is really small, so it's not a performance issue.
The grump-en and grump-dis scripts have been replaced by node scripts and can be executed through yarn, the documentation has been updated.
To configure you environment:
`composer update`
`yarn grump-en`

Should be enough, you should now see your commit "failing" if you modify any twig template introducing a violation (like a space at the end of a line, two spaces, whatever) while if you then fix it you should see the green output by grump-php plus the checks by husky.
When it fails it should show the output of the twigcs script, unfortunately in this version is not possible to limit the output to "errors" and we receive lots of warnings that are false positive.
I briefly tested the 4.0 version of twigcs which is still in beta, it spots apparently many violations in our templates, we might soon move to that where we will also be able to set the twig version to check the code against.

This solution is using grep, hope it's not an issue for the current developers.
The changes in the site header template are the result of this implementation, grump-php was "failing" due to that.

- [ ] I have provided tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ php/images
 yarn-error.log
 .idea
 composer.lock
+grump-report.txt
--- a/composer.json
+++ b/composer.json
@@ -13,13 +13,13 @@
     "grump-en": [
       "echo \"php ./vendor/bin/grumphp git:pre-commit\" >> .git/hooks/pre-commit"
     ],
-    "twig-cs": "./vendor/bin/twigcs ./src/*/packages",
+    "twig-cs": "./vendor/bin/twigcs \"./src/*/packages\"",
     "cs": "./vendor/bin/twigcs",
     "grump-cs": "php ./vendor/bin/grumphp run",
     "grump-dis": "./scripts/grump-dis.sh"
   },
   "require-dev": {
-    "friendsoftwig/twigcs": "^3.1.3",
-    "phpro/grumphp": "^0.16.1"
+    "friendsoftwig/twigcs": "^3.2.2",
+    "phpro/grumphp": "^0.18.0"
   }
 }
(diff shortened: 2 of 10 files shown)


Joosthe:
When i did a composer update I already get:
[ErrorException]                                           
 proc_open() expects parameter 1 to be string, array given

Joosthe:
But appart from the error it does seem to work fine although I do get al lot of false positives with the "WARNING Unused variable: ...". Which prevent me from commiting, But i thought you had a solution for this.

planctus:
Well, @Joosthe , if you had problems running composer update the results of this "test" is unpredictable, what i didn't get is whether this error you had went away and you were able to run composer update or not.

For the false positive is a known issue, but i don't know the reason why you had those reported, if i try to commit on this branch with no changes this is what i get:
image

Which is that grump-php checked all the templates, it succeded and husky run its script, there is nothing to commit so nothing happens, but the commit would succeed.

Now, if you had those warnings reported in the command line is because inside of that long list of warnings you probably had an "ERROR", i know, you have to look carefully to find which one is that.
And in this case it should have prevented you from committing (which is good), but in theory in this branch you don't have "errors" reported by twigcs so we need to understand. Could you try to find this "error" in the middle of the reported warnings?

papegaill:
> When i did a composer update I already get:
> [ErrorException]
> proc_open() expects parameter 1 to be string, array given

Added the same error, but this error has been around for a long time.
If you run composer update again, the error is no longer declared.
For the rest everything works fine on my side 👍

PR #410, opened 2020-04-17
feat(ecl-compliance): message, adding child template - FRONT-1093

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
+++ b/src/ec/packages/ec-component-button/__snapshots__/button.test.js.snap
@@ -205,7 +205,7 @@ exports[`EC - Button with validation enabled and missing input data returns the 
   <!-- ################################################################################################################
     ecl2-compliance -->
   <div
-    class="ecl2-compliance ecl-u-border-width-1 ecl-u-border-color-green ecl-u-border-style-solid ecl-u-ma-xs ecl-u-pr-s ecl-u-bg-white"
+    class="ecl2-compliance ecl-u-border-width-1 ecl-u-border-color-red ecl-u-border-style-solid ecl-u-ma-xs ecl-u-pr-s ecl-u-bg-white"
   >
     <p
       class="ecl-u-type-color-grey ecl-u-ml-m ecl-u-type-xs"
--- a/src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-message.html.twig
+++ b/src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-message.html.twig
@@ -1,0 +1,23 @@
+{#
+  Parameters:
+  - variant: (string),
+  - icon: (object),
+  - title: (string),
+  - close: (object)
+  - description (text)
+#}
+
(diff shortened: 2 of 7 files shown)


kalinchernev:
Checks in compositions provide mixed feedback which is not great in my personal opinion.
In this component you have more mandatory elements than compliance checks.
With that, I removed the label for the close button and got the following:
not_sure
Then as a user of this UI in storybook I'll be very confused if my thing is compliant or not as there are several feedback messages all around without coherency.

planctus:
I see @kalinchernev, i understand the general concern but in this case the confusion is due to a "bug" in this implementation, i think, basically the check on the label should be replicated also in the component validation because it invalidates that as well, although is not necessarily so clear what would be a correct implementation and what is not. 
In this case for instance you can assume that in the messages the label of the close button is mandatory because being a button it has to be there, but there is no a fully "explicit" mention of the label in https://ec.europa.eu/component-library/ec/components/message/usage/, for instance.

So, i will fix this, and then also the messages validation would fail when "removing" the label of the close button, not saying that we will never have potentially confusing output with the approach we have, but i'm rarely feeling dizzy abbout it, it is normally pretty clear what the checks are about, although we might want to "improve" the output we give.
At the moment is pretty basic and takes a bit of "room" in the output, we could define a more concise, and maybe differentiated output for the main component checks and the ones inherited from the basic components included.

I also think in general that this is a way to stress this "requirements" as provided by dg comm, potentially finding some conflict or more likely grey areas in them, we would then be able to collect these instances because sometimes it might be that the flaws on our side, when validating, could be due to an incorrect defintion of the specs, more than an issue in our logic. 
An example is the timeline, the usage page is completely wrong and we are still waiting for an update of that.
Thanks a lot for the review, @kalinchernev ;)

kalinchernev:
Few thoughts after reading https://github.com/ec-europa/ecl-twig/pull/410#issuecomment-621679481
- my comment was targeting the feedback provided by storybook know rather than "business" specifics, I honestly do not care who specifies what as required whenever or wherever
- if you could think of a way to provide central feedback in 1 message through src/ec/packages/ec-component-ecl-compliance/ecl-compliance.html.twig instead of 3 messages all over the component it will be a useful improvement

planctus:
You mean still delegating checks to the included templates but centralising the output so that each of the components validation output would be gathered in a single "div"..?
The thing is that like this the output for a specific included component is ideally always (there are cases where the output is broken or just not clear enough, though) close to the element it refers to, this is like a "visual debug" tool, in a way, if we centralise the output i'm not sure we would reach the goal, imho, there would be potentially lots of messages concentrated in one single container reporting the output of all the included components, imagine in a list of links, it would not help, i think.
At the same time we could simply drop the success message, this has always been a preference of alexis, for instance and i partially undesrtand it, but i believe that if you test something you want to see if this succeded or not, that's why you currently see it.
Anyway, as anything new, this is the first iteration on it and it's foreseen to find flaws and limits in this approach, therefore brainstorming on what could improve the current implementation is very, very welcome, thanks for this, @kalinchernev

PR #423, opened 2020-04-22
feat(ecl-compliance): Media Container, adding child template - FRONT-1096

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
+++ b/src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
@@ -1,0 +1,56 @@
+{#
+  Parameters:
+    alt (string),
+    image (string),
+    embedded_media: (string),
+    sources: (array of objects),
+    tracks: (array of objects)
+#}
+
+{% extends "@ecl-twig/ec-component-ecl-compliance/ecl-compliance.html.twig" %}
+
+{% set message = '' %}
+{% set component = 'media-container' %}
+{% set no_embedded_media_case = embedded_media is not defined or embedded_media is empty %}
+{% set no_sources_case = sources is not defined or sources is empty or sources is not iterable %}
+{% set no_tracks_case = tracks is not defined or tracks is empty or tracks is not iterable %}
+{% set no_video_resources_case = no_sources_case and no_tracks_case %}
+{% set no_image_case = image is not defined or image is empty %}
+{% set no_alt_case = no_image_case == false and (alt is not defined or alt is empty) %}
+
+{# no media provided at all #}
+{% set not_compliant = no_image_case and no_embedded_media_case and no_video_resources_case %}
+{% if not_compliant %}
+  {% set message = 'One media is required for the media-container component, an "image", an "embedded_media" or video "tracks" and "sources".' %}
+{% endif %}
+{# No embedded_media, we check for video sources #}
+{% if no_embedded_media_case and no_sources_case == false %}
+  {% for source in sources %}
+    {% if (source.src is not defined or source.src is empty) or (source.type is not defined or source.type is empty) %}
+      {% set not_compliant = true %}
+    {% endif %}
+  {% endfor %}
+  {% if not_compliant %}
+    {% set message = message ~ 'The "sources" parameter should be an array with a "src" and a "type" for each element.' %}
+  {% endif %}
+{% endif %}
+{# No embedded_media, we check for video tracks #}
+{% if no_embedded_media_case and no_tracks_case == false %}
+  {% for track in tracks %}
+    {% if (track.kind is not defined or track.kind is empty) or
+      (track.label is not defined or track.label is empty) or
+      (track.src is not defined or track.src is empty) or
+      (track.src_lang is not defined or track.src_lang is empty) %}
+      {% set not_compliant = true %}
+      {% set tracks_issue = true %}
+    {% endif %}
+  {% endfor %}
+  {% if not_compliant and tracks_issue is defined %}
+    {% set message = message ~ ' The "tracks" parameter should be an array with a "src", a "kind", a "label" and a "src_lang" for each element.' %}
+  {% endif %}
+{% endif %}
+{# No embedded_media, no video, we check for a missing alt attribute in the image #}
+{% if no_embedded_media_case and no_video_resources_case and no_image_case == false and no_alt_case %}
+  {% set not_compliant = true %}
+  {% set message = 'If an image is set as the media also the "alt" parameter needs to be set.' %}
+{% endif %}
--- a/src/ec/packages/ec-component-media-container/README.md
+++ b/src/ec/packages/ec-component-media-container/README.md
@@ -25,7 +25,7 @@ npm install --save @ecl-twig/ec-component-media-container
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
   - "name" (string) Attribute name, eg. 'data-test'
   - "value" (string) Attribute value, eg: 'data-test-1'.
-- **"_compliance_"** (boolean) (default: false)
+- **"_compliance_"** (boolean) (default: false) Activates debug
 
 ### Blocks:
 
--- a/src/ec/packages/ec-component-media-container/README.md
+++ b/src/ec/packages/ec-component-media-container/README.md
@@ -25,6 +25,7 @@ npm install --save @ecl-twig/ec-component-media-container
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
   - "name" (string) Attribute name, eg. 'data-test'
   - "value" (string) Attribute value, eg: 'data-test-1'.
+- **"_compliance_"** (boolean) (default: false)
 
 ### Blocks:
 
--- a/src/ec/packages/ec-component-media-container/__snapshots__/media-container.test.js.snap
+++ b/src/ec/packages/ec-component-media-container/__snapshots__/media-container.test.js.snap
@@ -125,3 +125,36 @@ exports[`EC Media Container Media Container generic tests renders correctly with
   </figure>
 </test>
 `;
+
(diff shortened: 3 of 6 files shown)


src/ec/packages/ec-component-media-container/README.md
```diff
@@ -25,6 +25,7 @@ npm install --save @ecl-twig/ec-component-media-container
 - **"extra_attributes"** (optional) (array) (default: []) Extra attributes
   - "name" (string) Attribute name, eg. 'data-test'
   - "value" (string) Attribute value, eg: 'data-test-1'.
+- **"_compliance_"** (boolean) (default: false)
```
papegaill, line 28: Maybe we can have a small description like other component?
"compliance" (boolean) (default: false) Activates debug.

planctus: sure ;)

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% set component = 'media-container' %}
+{% set no_embedded_media_case = embedded_media is not defined or embedded_media is empty %}
+{% set no_sources_case = sources is not defined or sources is empty or sources is not iterable %}
+{% set no_tracks_case = tracks is not defined or tracks is empty or tracks is not iterable %}
+{% set no_video_resources_case = no_sources_case and no_tracks_case %}
+{% set no_image_case = image is not defined or image is empty %}
+{% set no_alt_case = no_image_case == false and (alt is not defined or alt is empty) %}
+
+{# no media provided at all #}
+{% set not_compliant = no_image_case and no_embedded_media_case and no_video_resources_case %}
```
kalinchernev, line 22: It's very possible I'm not getting logic of cascading AND statements to catch possible wrong scenarios, though to be honest I'd expect OR statements combination.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% if not_compliant %}
+  {% set message = 'One media is required for the media-container component, an "image", an "embedded_media" or video "tracks" and "sources".' %}
+{% endif %}
+{# No embedded_media, we check for video sources #}
+{% if no_embedded_media_case and no_sources_case == false %}
```
kalinchernev, line 27: same question whether an OR wouldn't be more appropriate in order to maximize catching errors.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
planctus: I can explain, this is a particular use case, the logic is always tailored to what the template is doing, and sometimes it might be that is the logic in the template is not so strong or straightforward, in general.
Anyway, here the thing is that only when all of these use cases are true the component is not valid, that's why, exceptionally, in this component the not_compliant is a concatenation of ANDS.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+  {% for source in sources %}
+    {% if (source.src is not defined or source.src is empty) or (source.type is not defined or source.type is empty) %}
+      {% set not_compliant = true %}
+    {% endif %}
+  {% endfor %}
+  {% if not_compliant %}
+    {% set message = message ~ 'The "sources" parameter should be an array with a "src" and a "type" for each element.' %}
```
kalinchernev, line 34: As you are iterating through `sources` for the compatibility check, you can (I think) allow yourself to use this `message` variable to include more specific information which source exactly is the problematic.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
planctus: The comments here were trying to help, we follow the logic in the template, which renders the embedded_media if it's present, then it goes rendering a video with sources and tracks.
In this case the condition needs to check for the embedded_media before even checking for the sources, and only in case the first is not set and the second is, it loops into the sources to spot some missing parameter.
So, no, in this case an OR approach would not catch more errors, it would instead create errors where we don't have, potentially, imagine you set the embedded_media but then you also have a sources array defined, with some problems, the template renders the embedded_media and it's correct, valid, although it contains some additional, unneeded data.
Hope this explains :)

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+  {% for track in tracks %}
+    {% if (track.kind is not defined or track.kind is empty) or
+      (track.label is not defined or track.label is empty) or
+      (track.src is not defined or track.src is empty) or
+      (track.src_lang is not defined or track.src_lang is empty) %}
+      {% set not_compliant = true %}
+      {% set tracks_issue = true %}
+    {% endif %}
+  {% endfor %}
+  {% if not_compliant and tracks_issue is defined %}
```
kalinchernev, line 48: Similar comment as in https://github.com/ec-europa/ecl-twig/pull/423/files#r421329856
Changing the `not_compliant` marker inside a loop allows you for adding more specific information which iteratee is wrong exactly.
It's also not clear to me why there's the additional `tracks_issue` here when it's not used for providing additional context.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
planctus: it's true, we can clearly be more granular, but the evaluation i made is that this was the right "balance", between the logic to be set, the messages to be prepared and the "value" in term of information provided to the user. Clearly we could also specifically mention the element or th elements in a loop that are causing the failure, but this is a tool for developers, we aim to help, but i think it's enough if you see that your problem is coming from a certain array where we tell you which are the needed keys.
In this case for instance you have sources but also tracks, the number of different use cases would increase a lot, that's the reason for this "limitation".
Anyway, this approach has been used many times, in many loops, therefore i would in case considering a second iteration on this, if we want to, my personal opinion is that is not worth but let's decide it together :)

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
planctus: `tracks_issue` is there because `not_compliant` might be true from the previous loop (sources), therefore we need and additional flag to understand whether to add the messages about the tracks or not.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
kalinchernev: got the idea

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% set no_sources_case = sources is not defined or sources is empty or sources is not iterable %}
+{% set no_tracks_case = tracks is not defined or tracks is empty or tracks is not iterable %}
+{% set no_video_resources_case = no_sources_case and no_tracks_case %}
+{% set no_image_case = image is not defined or image is empty %}
+{% set no_alt_case = no_image_case == false and (alt is not defined or alt is empty) %}
+
+{# no media provided at all #}
+{% set not_compliant = no_image_case and no_embedded_media_case and no_video_resources_case %}
+{% if not_compliant %}
+  {% set message = 'One media is required for the media-container component, an "image", an "embedded_media" or video "tracks" and "sources".' %}
```
kalinchernev, line 24: Trying out the compliance checker component from the knobs in storybook. Getting a generalized message doesn't give as useful an insight as it's aiming I think. When I'm on an image media container I'd expect to get a suggestion only for the image-related component with its required data structure.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% endif %}
+{# No embedded_media, we check for video sources #}
+{% if no_embedded_media_case and no_sources_case == false %}
+  {% for source in sources %}
```
kalinchernev, line 28: About this variable, it's quite unfriendly to have this in a text field

```
[
  {
    "src": "https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/720/Big_Buck_Bunny_720_10s_2MB.mp4",
    "type": "video/mp4"
  },
  {
    "src": "https://test-videos.co.uk/vids/bigbuckbunny/webm/vp9/720/Big_Buck_Bunny_720_10s_2MB.webm",
    "type": "video/webm"
  }
]
```

Isn't it the idea of storybook knobs to be user-friendly way to work with components?

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
kalinchernev: `this is a tool for developers` if this is true, why not making a CLI for these checks?

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-media-container.html.twig
kalinchernev: clearly you can have different variables for the separate complience checks then instead of working around this.

PR #435, opened 2020-04-27
feat(ecl-compliance): Card, adding child template - FRONT-1058

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-card/card.story.js
```diff
@@ -90,27 +92,27 @@ const formatCard = data => {
...
   // Meta.
   const meta = data.meta
-    ? array('card.meta (tags)', data.meta, '|', optionalGroupId)
+    ? array('card.meta', data.meta, '|', tabLabels.optional)
     : false;
 
   const formatTag = (tag, index) => {
     tag.label = text(
-      `card.tags.label (Tag ${index})`,
+      `card.tags[${index}].label`,
```
kalinchernev, line 113: Why making the index a string here? It doesn't seem necessary.

planctus: To get this, no? 
image
we use the same also for infos and links

planctus:
> "and none of" part is not a clear message.

I really welcome any proposal in the way synctatically or semantically write this messages, sometimes it's not easy and this one in link was the hardest.

The fact that you get the warning message from the link component is correct in the current approach, while implementing the card that way you are also wrongly implementing a "link" and you get "potentially" additional info from that component.
The one i expressed with a certain difficulty, is indeed the only condition where we define the link as not compliant. The logic behind that is that when you implement a link you could have an empty label, for instance, but then we would expect an icon, maybe it's too verbose the way it is, but i thought it needed these info to be understood, let's make it better if it's not clear.

src/ec/packages/ec-component-card/card.story.js
kalinchernev: oh, sorry, it's the whole link, it's correct!

kalinchernev:
> while implementing the card that way you are also wrongly implementing a "link" and you get "potentially" additional info from that component.
I haven't implemented anything here

The suggestion for making the pattern more clear is being already discussed at https://github.com/ec-europa/ecl-twig/pull/437#discussion_r425091653

PR #437, opened 2020-04-28
feat(ecl-compliance): Page Banner, show/hide buttons, adding child template - FRONT-1081

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-page-banner/page-banner.story.js
```diff
@@ -1,65 +1,161 @@
...
+// Preserve the adapted specs.
+let defaultData = { ...bannerDataDefault };
+let imageData = { ...bannerDataImage };
+let imageShadeData = { ...bannerDataImageShade };
+let primaryData = { ...bannerDataPrimary };
+let leftData = { ...bannerDataAlignLeft };
+
+// Show/hide buttons for optional elements.
+const defaultDescBtnToggler = () => {
+  defaultData.baseline = defaultData.baseline
```
kalinchernev, line 36: It's the first time I see boolean false on the place of ternary true expression statement. Hard to read and follow.
Please find a way to rewrite to have the conventional `condition ? exprIfTrue : exprIfFalse` or `||` as `defaultData.baseline = bannerDataDefault.baseline || defaultData.baseline;`

src/ec/packages/ec-component-page-banner/page-banner.story.js
```diff
@@ -68,22 +164,89 @@ storiesOf('Components/Banners/Page Banner', module)
...
+        'With or without title',
+        imageShadeTitleBtnToggler,
+        tabLabels.cases
+      );
+      button(
+        'With or without baseline',
+        imageShadeDescBtnToggler,
+        tabLabels.cases
+      );
+      button('With or without cta', imageShadeCtaBtnToggler, tabLabels.cases);
```
kalinchernev, line 194: "cta" by itself doesn't mean much. Could be CTA, or cta link, cta button or whatever, otherwise no context.

src/ec/packages/ec-component-ecl-compliance/components/ecl-compliance-page-banner.html.twig
```diff
@@ -0,0 +1,23 @@
...
+{% set component = 'page-banner' %}
+{% set no_image_case = false %}
+
+{% if type in ['image', 'image-shade'] %}
+  {% set no_image_case = image is not defined or image is empty %}
+{% endif %}
+
+{% set not_compliant = no_image_case %}
+
+{% set message = message ~ ' when type is "' ~ type ~ '" "image" is mandatory.' %}
```
kalinchernev, line 19: `[page-banner] when type is "image" "image" is mandatory.` this message is unclear.

src/ec/packages/ec-component-page-banner/page-banner.story.js
```diff
@@ -1,65 +1,161 @@
...
   }
 
   data.type = select('type', [data.type], data.type, tabLabels.required);
-  data.title = text('title', data.title, tabLabels.required);
+  if (data.title) {
+    data.title = text('title', data.title, tabLabels.optional);
+  }
   if (data.baseline) {
-    data.baseline = text('baseline', data.baseline, tabLabels.required);
+    data.baseline = text('baseline', data.baseline, tabLabels.optional);
```
kalinchernev, line 116: What property is optional and what is required is a mystery. How do you decide this?
Group of mandatory for primary, default and alight-left knobs don't make sense to me. We had had a chat about this in slack, but here are my thoughts again:
- having a dropdown select in 1 element doesn't make sense. If there is 1 value, it's not an option, no need to be in the storybook UI.
- having a dropdown with a string being an image path doesn't make sense. If it's a text field which can be modified, then it's a real knob and it will have a value of existence in the storybook UI.
- displaying type knob in this case is unnecessary. If there was 1 view where the UI can really change based on this type, ok, but this is not the case.

src/ec/packages/ec-component-page-banner/__snapshots__/page-banner.test.js.snap
```diff
@@ -441,6 +441,95 @@ exports[`EC - Page Banner image renders correctly with extra class names 1`] = `
...
+  <div
+    class="ecl2-compliance ecl-u-border-width-1 ecl-u-border-color-red ecl-u-border-style-solid ecl-u-ma-xs ecl-u-pr-s ecl-u-bg-white"
+  >
+    <p
+      class="ecl-u-type-color-grey ecl-u-ml-m ecl-u-type-xs"
+    >
+      <b>
+        [page-banner]
+      </b>
+        when type is "image" "image" is mandatory.
```
kalinchernev, line 527: Commented in the other lines of code, here is the suggestion: "when type of [which component] is "image", "image" property of [which component] is mandatory."

src/ec/packages/ec-component-page-banner/page-banner.story.js
kalinchernev: I'm still trying to understand the logic here, really hard.
Do you try to return false because of button handlers? Reading docs of the addon
```
Button knobs cause the story to re-render after the handler fires. You can prevent this by having the handler return false.
```

Do you try to set false for a property which is managed by a boolean knob otherwise?

planctus: There might be a better way for doing this, but this are toggler buttons, the way they act is to remove a piece of data, the false it's set to hide the "element" when the data is in place, while the copied data from the cloned object is used to restore it when the user clicks again.
I refactored it with ifs to avoid confusion, deleting the piece of data instead of setting in to false.

Buttons in twig are used in the "use cases" tab, to quickly show the different look and feel of a component when it's relevant, normally we use them for the big ones, and this one seemed relevant due to the fact that there is nothing mandatory, basically. 
It's true that they overlap the work by some knob and they don't play well together, though.

src/ec/packages/ec-component-page-banner/page-banner.story.js
planctus: i renamed it to "call to action"

src/ec/packages/ec-component-page-banner/page-banner.story.js
planctus: > What property is optional and what is required is a mystery. How do you decide this

Well, the primary source is the ECL usage page of a component, i'm not deciding myself what is optional or mandatory, ther are grey areas, though, and i think this is also a good way to discover them, sometimes other considerations have been used to compensate these grey areas, for this i use the W3c for references around the validity of html elements.

`We had had a chat about this in slack, but here are my thoughts again:`

But i haven't forgotten your remarks, @kalinchernev, i tried to explain the reason why of these choices and i cannot necessarily convince you about them but i cannot even assume that a set of consistent choices has to be changed because of your remarks.

Most of the components use icons, for instance, and it's crucial in this case to pass the icon path to the component to make it work, so if we wanted to follow all the data through the knobs, the icon_path should not be skipped and having it as a text field would just give the users the chance to breake the demo.

Also the `type` is an example of something we already encountered, it controls the variant of the component, it's an important piece of data to be shown but we have different demos for all the variants, so if we give the chance to select the other ones the user would basically reproduce an already existing example in the demo, just that.

Having said this, it would come easy to turn the icon_path, or logo_path into textfields and the variants to be selectable, if that's what you want, but i don't think it would be better, i wouldn't have looked for this solution ottherwise.

src/ec/packages/ec-component-page-banner/__snapshots__/page-banner.test.js.snap
planctus: now it looks like this:
`[page-banner] When type is "image" the "image" key needs to be set and filled in.`
your one would be:
`[page-banner] When type of page-banner is "image", "image" property of page-banner is mandatory`
 
not too different, the thing is that we are "prefixing" every block of compliance with the name of the component, including the variant name, i wouldn't repeat it also in the message.
I'll use property instead of key, though ;)

kalinchernev: > now it looks like this:
[page-banner] When type is "image" the "image" key needs to be set and filled in.

Not true, the message is
> [page-banner] when type is "image" "image" is mandatory.

And it's obviously not descriptive for what the comment came about.

planctus:
I noticed that the variant was not passed to the parent template, fixed it, so now the message looks like this:
`[page-banner--image-shade] When type is "image-shade", the "image" property needs to be set and filled in.`

src/ec/packages/ec-component-page-banner/page-banner.story.js
kalinchernev: I've created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1159 
I am starting to get the feeling knobs are so sophisticated part of the ecl-twig project that the components are a minor detail around the core logic of displaying many knobs on each property regardless needs.

kalinchernev:
I created tickets for the points which can be improved later to not block the pull request.
There is something wrong with builds, please check before merging and moving forward.

planctus:
restarted the builds, drone issues

PR #441, opened 2020-05-04
feat(ecl-compliance): Table, adding child template - FRONT-1078

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-table/__snapshots__/table.test.js.snap
```diff
@@ -1102,6 +1102,216 @@ exports[`EC - Table Default renders correctly with row extra classes 1`] = `
...
+        <td
+          class="ecl-table__cell"
+          data-ecl-table-header="Type of contract"
+        >
+          Seconded National Expert (SNE)
+        </td>
+      </tr>
+    </tbody>
+  </table>
+  <!-- ################################################################################################################
```
kalinchernev, line 1298: Looking at the snapshot (not only this one, but this is specifically too big), only this part is actually useful for the assertion https://github.com/ec-europa/ecl-twig/pull/441/files#diff-516d297d2ac695f7864b6efaf0e8e1aeR1298-R1311

If you find a way to either assert only for message string value or this particular dom element it will improve the focus and usability of this assertion.

src/ec/packages/ec-component-table/table.test.js
```diff
@@ -70,6 +70,15 @@ describe('EC - Table', () => {
 
       return expect(render(withRowExtraClasses)).resolves.toMatchSnapshot();
     });
+
+    test('with missing input data and debug enabled shows the right warning message.', () => {
```
kalinchernev, line 74: You have written logic for 2 main cases: 1) wrong headers 2) wrong rows
Only the first case is covered.

planctus: yep, it's actually three cases, i can only test two of them in one shot, i added the second :)

PR #470, opened 2020-06-02
feat(ecl-compliance): Inner/outer report - FRONT-1150

planctus:
# PR description

The changes contained in this pr reflect an attempt to address many concerns raised during the work for the "compliance" track.
Some of these concerns are strictly related to this implementation, but some others involved existing implementations, so although the scope of this ticket was "just" to improve the reports offered by the compliance tool the work done here collected also other inputs and put them together.  Sorry for the amount of changes here, but it's still possible to clearly identify the changes made, in a pattern:

- A new parameter "inner_check" has been introduced, it is used to discriminate between the main scope of the "test" and the subsequent checks we perform on the included templates. So this is mainly an "internal" parameter, stil it's possible to instruct the tool to consider also some "main" components as "inner_check"(s).

- The report is now based on an hyper compact representation of the internal checks (an icon, where the title attribute holds the textual info) when this is possible, we need the path to the icons to be able to use it.

- In some cases, since it seemed reasonable to get this parameter as once more than requesting it multiple times when multiple icons are involved, a new parameter "icon_path" has been added to the components, in a soft way, as an optional parameter. If it's used then the path to the icon file doesn't have to be provided with all the icons implementation in a component.

- Most of the times we were using selects with single values to represent the path to the icons file or the logo. In these cases a "none" option has been added to give the option to interact with the knob and to test also the resulting "compliance" status.

- buttons have been removed from the card component, they were completely overlapping the work by the knobs and it didn't make sense. This simplified and fixed also the story, where now the notes tab is offering the complete input passed to the component.

- Strings for the compliance have been rewritten, the double quotes were not compatible with the "new" rendering in a title tooltip, they have been replaced by {{ }}

- The expression "the property needs to be set and filled in" has been preferred over "the property is mandatory" in order to specify that we expect a non empty value, not just a key.

- The html comment on top of the "main" check  is now reporting the "identifier", component + variant

- Many fixes have been applied in different contexts, while reviewing the components some of them popped out, especially about the "compliance" templates. 

## QA Checklist

- [ ] I have provided tests


src/ec/.storybook/preview-head.html
```diff
@@ -25,11 +30,50 @@
 ></script>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.8.3/polyfill.min.js"></script>
 <script src="https://unpkg.com/svg4everybody@2.1.9/dist/svg4everybody.min.js"></script>
+<script src="https://unpkg.com/@popperjs/core@2"></script>
+<script src="https://unpkg.com/tippy.js@6"></script>
```
papegaill, line 34: I like the new way of visualising internal errors with the tooltips, however, since the components are destined to be used outside of storybook, how will this react in a third-party application? like before?

src/ec/packages/ec-component-breadcrumb-core/ecl-breadcrumb-core.html.twig
```diff
@@ -21,11 +21,14 @@
         ...
       ]
     - "_compliance" (boolean) (default: false) Activates debug
+    - "inner_check" (boolean) (default: false) Inline compliance report
```
papegaill, line 24: I am not convinced of the name of this parameter. why did you put undescore to the **_compliance_** parameter but not this one? Isn't it better to name it **_compliance_inner_check_**?

src/ec/packages/ec-component-breadcrumb-core/ecl-breadcrumb-core.html.twig
```diff
@@ -21,11 +21,14 @@
         ...
       ]
     - "_compliance" (boolean) (default: false) Activates debug
```
papegaill, line 23: sometimes it's **__compliance__** with double underscore sometime not?

src/ec/packages/ec-component-breadcrumb-core/ecl-breadcrumb-core.html.twig
papegaill: Would it be nice to give a little more explanation on this parameter? Or update documentation as you describe the parameter in your PR description?

src/ec/.storybook/preview-head.html
papegaill: Little idea when checking the PR .. would be to have a system like drupal theme-debug, ie that most of the issues or messages becomes html comments in code via inspect .. a visual icon remain to inspect them..

src/ec/packages/ec-component-breadcrumb-core/ecl-breadcrumb-core.html.twig
planctus: The thing is that inner_check is basically an internal parameter because it's just the internal logic that sets it for the included templates.  The `_compliance_` one is the only relevant parameter from the user input data, `inner_check` is just offering a compact report if it's possible and it's mainly us deciding when to apply it, is not meant to be provided by the user.
I will add some documentation for us, the developers, then, about the name i don't know, i think i just wanted something "compact", but i understand that like this is not "connected" to the functionality, semantically speaking. I can rename it like you suggest.

src/ec/packages/ec-component-breadcrumb-core/ecl-breadcrumb-core.html.twig
planctus: it's a special parameter, that's why this idea of using this syntax with the "double" underscore , to make it stand out, anyway the parameter is `_compliance_`, i need to correct this.

src/ec/.storybook/preview-head.html
planctus: So, the inner_check uses a standard "title" attribute. This fancy tooltips are only for ecl-twig, to improve the demo, but it would work in any environment, just using the standard overlay handled by the browser.
I'm not sure i get the idea in the second part of your sentence.

docs/getting-started.md
```diff
@@ -31,9 +31,25 @@ There's also an extra package: `@ecl-twig/{ec,eu}-components` which is used to a
...
+
+### ECL Compliance
+
+The ecl-twig library comes with a special component named `ecl-compliance`, which defines a parent template for the ecl compliance checks to be performed, it also contains all the children templates for the checks applied to the single components.
+Each of our templates should have a corresponding child template defined so to make a compliance check available through our storybook demo and to all the consumers of the library thanks to the ecl-compliance package and a parameter applied to the component instantiation: `_ecl_compliance_`.
+
+An additional parameter `_ecl_compliance_inner_check_` is available for identifying the inner checks run by a component on all its included templates from the overall compliance performed on a component. This offers us the possibility of differentiating the output of the report. Mind that for this parameter to work also the `icon_path` needs to be available in order to render an icon for the success or the failure of the checks.
```
kalinchernev, line 47: 1) low-value UI gimmicky (icon) imposes a requirement/dependency
2) popular flag is `verbose` and not `inner`. Yes, it's wording, but rarely anyone cares about inner work until module the software works. "Verbose" suggests that the user is debugging a reason something is not working and is ready to see the lower-level logic.

The sentence is of course too long and can be broken down.

docs/getting-started.md
```diff
@@ -31,9 +31,25 @@ There's also an extra package: `@ecl-twig/{ec,eu}-components` which is used to a
...
+
+The file name of the child template is expected to follow this convention:
+_ecl-compliance-{nameOfTheComponent}.html-twig_
```
kalinchernev, line 50: In a similar way a react template is created for demo during work done on a component in ECL and this is seen as overcomplication, ecl-twig is growing in terms of information encoded in components' templates. Tricky and proving that any project growing adds code to cover some purpose, but don't be surprised if someone tells you that complience checker templates in addition to components templates is over-complication.

The tricky part is not so much about writing loads of code with checks, but maintaining it. Especially having it in twig language which is not the most interoperable.

docs/getting-started.md
```diff
@@ -31,9 +31,25 @@ There's also an extra package: `@ecl-twig/{ec,eu}-components` which is used to a
...
 
-You can use knobs to make the showcase more dynamic and interactive.
+In ecl-twig the storybook knobs are extensively used to accurately represent the full input data with the aim of offering the possibility to customise the component and retrieve the correct object to the passed to the component in the `notes` tab.
+Their label is supposed to be defined the same way the property is accessible in the expected input data.
+Ex: `link.icon.path`
 
-More features will come in the near future...
+Knobs in ecl-twig are grouped together to simplify their usage and offer additional information, like whether the property controlled by the knob is mandatory or optional, for instance, the available groups are _mandatory_, _optional_, _use cases_, _status_
+In ecl-twig buttons knobs are used to defined different use cases the component might have, when blocks of information are optional and can be toggled on and off in a component demo.
+In many cases (like icons, path to the logo...) there is no real advantage for the user in editing the value used in the demo, therefore we only offer the possibility of unsetting it, mainly for testing the compliance checks we perform.
```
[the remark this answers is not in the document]
planctus: Well, didn't really expect you to embrace this solution, maybe, but i hope it takes into account the strongest argument against the previous implementation, where a knob was defined as non interactive element, basically.
This way we offer the only interaction that makes sense, un-setting the value of the "path" to whatever resource we point to, for then being able to restore it. 
In a text field, if the user deletes the current value to test the results he would not be able to retrieve it back, instead.

docs/getting-started.md
planctus: 1) it's optional, if no `icon_path` is available it renders the standard output. `_compliance_inner_check_` It's also an internal parameter for the most, we know when we have an icon path available and when not, basically, the docs here are for the developer awareness of when it's possible to apply an "inner" report in the compliance checks added to a component he might be working on.

2) in this case verbose would just mean the opposite, here the idea is to make it more compact, instead, when we know it's an "automatic" check performed by the internal logic of the templates, the final users can simply ignore it, not much to do with it.

docs/getting-started.md
planctus: I find the comparison with the situation in ECL you mentioned totally arbitrary, @kalinchernev, not following it, then.
it's not really surprising indeed to deal with some additional code in implementing a new functionality, what might be surprising is maybe to consider this an overhead in this case.
 
In my experience, for instance, the real overhead in working on this library was the total lack of confidence the developers had about the "quality" of the components they were implementing, because there was not much there to help them.
Now we have many pillars where to base our confidence on, instead, and yes, each of them required work and code, also extensively adding knobs to all the stories was an overhead, but the additional value of it justifies it, if the users are now capable of seeing and editing the whole data structure and retrieve the right include code after they customised it.

In this case the added value is both for us, as developers and for the final users, some twig logic in a template doesn't sound as an extraordinary effort on our side, it's a twig templates library, nothing really looks tricky or complicated, to be honest.
About the "maintenance", this would only be an issue if the existing components would run into multiple iterations, which doesn't seem the case in a library that should be close to its final state and anyway i hardly see how this would be "critical", it's just an additional task for the developer, but it's for the good.

planctus:
> - messages are still displayed within components and not in a single report
> 
> - the addition of documentation is appreciated, please try to break down long sentences into smaller ones to make them easier to follow
> 
> -  try to find a way to take checkers logic out of twig-specific logic if you can, as the direction of the compliance checker is growing in direction to be hard to maintain


- I never meant to put all the reports from the single components in a single one, it was your preference and i would have considered that as an alternative if we had it available now.

- I did work on the sentences in the docs, should be more compact now, although i'm not necessarily having issues with a more verbose approach. 

- I mentioned a few times that this is an implementation using only twig because it is meant to be used by any of the consumers the same way we do, hence the value of using only twig, here i read many times concerns about this specifically, which are legitimate, but they clearly invest the foundation of this implementation. 

As i see it, since i experienced a progressive decrease in enthusiasm and motivations on this functionality while going through all the latest work performed, i feel that we can drop this whole work, it was good to have it an epic, from this point of view, although we also performed additional operations on these prs to tackle other things, but no reason why releasing something no one believes in.

docs/decisions/003-ecl-compliance.md
```diff
@@ -0,0 +1,121 @@
...
+| **Proposed**  | 2020-04-30                                                     |
+| **Accepted**  | 2020-06-19                                                     |
+| **Driver**    | @planctus                                                      |
+| **Approver**  | @kalinchernev, @papegaill                                      |
+| **Consulted** | @papegaill, @emeryro, @kalinchernev, @pgingao                  |
+| **Informed**  | consumers                                                      |
+
+## Decision
+
+In order to facilitate the implementation of the ECL components while also enforcing somehow the correct usage of those, the ecl-twig library offers an ecl-compliance tool in the form of checks that can be performed on a component to test whether it respects the editorial guidelines defined for it.
```
kalinchernev, line 14: `the ECL components` => `ECL components`

kalinchernev: somehow can be removed

kalinchernev:
```suggestion
In order to facilitate and enforce correct implementation of ECL components in Twig-based templates the ECL Twig library provides `ecl-compliance` tool. This tool is a collection of Twig template providing an extended list of business and technical requirements expected from a component's implementation. 
```

docs/decisions/003-ecl-compliance.md
```diff
@@ -0,0 +1,121 @@
...
+This tool uses only logic defined in twig templates in order to make the functionality available to any consumer of the library without any additional need, in a transparent way.
+
+## Context
+
+Ecl twig is a library of twig templates implementing vanilla ECL components: https://ec.europa.eu/component-library/.
```
kalinchernev, line 19: `Ecl` -> ECL, or something else which can make naming of the project consistent throughout the documents within the repository.

planctus:
Thanks @kalinchernev again for the review, i improved a bit these two pages, i hope ;)

PR #477, opened 2020-06-12
feat(gallery): Component and resources updates for v2.29.0 - FRONT-1222

planctus:
# PR description

Component aligned with ecl 2.29.0.

It is possible to retrieve the diff between our rendered html and the ecl one:
`yarn check:component ec gallery`
`yarn ecl-diff ec`, 
choose gallery, all the other options are already fine, just skip them.


planctus:
Thanks @emeryro ..fixed it ;)

PR #485, opened 2020-07-01
feat(knobs): Replacing select elements with optionsKnob in multiple stories - FRONT-1253 

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-tag/tag.story.js
+++ b/src/ec/packages/ec-component-tag/tag.story.js
@@ -34,7 +34,7 @@ const prepareTag = (data, link, aria) => {
       tabLabels.required
     );
   }
-  if (data.default_icon_path) {
+  if (aria) {
     data.default_icon_path = optionsKnob(
       'default_icon_path',
       { current: defaultSprite, 'no path': '' },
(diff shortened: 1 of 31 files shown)


planctus:
thanks @kalinchernev for the review, i'll check all these use cases

planctus:
Thanks a lot @kalinchernev for the reivew:

- Some changes were lost when i recreated this pr, added them back and applied the chance also to the story-utils package

- accordion was a bad implementation, i just reverted the id knob to be a standard text field.

- The multiple contexts where the icons get removed by toggling the icon_path in language list overlay is because we use only one parameter, so it's correct.

- The cases where you said the toggler did not have effect i could not reproduce them, for the social components it might be that the icon you were resetting was the one for the hover? If that's the case i appended (icon hover) to the name's knob for making it clear. For the menu legacy is see the icon removed when we toggle the path.

src/ec/packages/ec-component-tag/tag.story.js
```diff
@@ -35,10 +35,11 @@ const prepareTag = (data, link, aria) => {
     );
   }
   if (data.default_icon_path) {
-    data.default_icon_path = select(
+    data.default_icon_path = optionsKnob(
       'default_icon_path',
-      [defaultSprite],
+      { current: defaultSprite, 'no path': '' },
```
kalinchernev, line 40: This one doesn't look quite right on toggling
disappearing-knob

planctus:
Thanks again for the review, @kalinchernev, i fixed the behaviour in the tag component ;)

PR #490, opened 2020-07-02
feat(scripts): Removing legacy scripts from package.json - FRONT-1257

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


package.json
```diff
@@ -45,12 +45,7 @@
     "twing:copy": "ncp static build",
     "twing:clean": "rimraf build dist",
     "twing:dist": "npm-run-all twing:clean storybook-build-twing:* twing:copy",
-    "update-version": "lerna version --no-push --no-git-tag-version",
-    "clean": "yarn twing:clean",
```
kalinchernev, line 49: Having a central `clean` which runs `run-p` of both clean for php and twing will still be useful.

package.json
```diff
@@ -45,12 +45,7 @@
...
-    "copy": "yarn twing:copy",
-    "dist": "yarn twing:dist",
```
kalinchernev, line 51: Also one central `dist` running both systems' dist will also be useful. Having `yarn clean && yarn dist` will prepare everything in one line clean line

planctus:
Thanks for the review @kalinchernev, i've tried to take into account all the remarks and i've tested them, in particular changing the way we called render.js requested a few adaptations, but now should be all fine.
I see that now the build of EU is always failing on the digit drone, it simply never happened in all the builds i tried locally and with these latest changes i see again also php failing with the segmentation fault, which is something i could reproduce locally when we made the first bunch of changes in the package.json and i was trying to do part of the things i did today (running everything with lerna from php-storybook, basically), so i am bit worried :)

PR #499, opened 2020-07-15
feat(site-headers): EU, adapting templates, adding demos - front-1285

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-site-header-standardised/__snapshots__/site-header-standardised.test.js.snap
```diff
@@ -13,17 +14,9 @@ exports[`EC - Site Header Standardised Default renders correctly 1`] = `
...
-        >
-          <img
-            alt="European Commission logo"
-            class="ecl-site-header-standardised__logo-image"
-            src="/logo--en.svg"
-            title="European Commission"
-          />
-        </a>
+          href=""
+        />
```
papegaill, line 19: no more link, aria-label here seems strange.

src/ec/packages/ec-component-site-header/__snapshots__/site-header.test.js.snap
```diff
@@ -21,7 +21,7 @@ exports[`EC - Site Header Default renders correctly 1`] = `
           <img
             alt="European Commission logo"
             class="ecl-site-header__logo-image"
-            src="/logo--en.svg"
+            src=""
```
papegaill, line 24: same here no logo path seems strange.

planctus: Fixed it ;)

src/ec/packages/ec-component-site-header-core/README.md
```diff
@@ -14,7 +14,8 @@ npm install --save @ecl-twig/ec-component-site-header-core
   - "title" (string) (default: ''): Logo title attribute.
   - "alt" (string) (default: ''): Logo alt attribute.
   - "href" (string) (default: ''): Logo URL.
-  - "src" (string) (default: ''): Logo image file path, eg. dist/images/logo--en.svg.
+  - "src_desktop": (string) (default: ''): Desktop logo image file path
+  - "src_mobile": (string) (default: ''): Mobile logo image file path
```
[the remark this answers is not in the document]
planctus: Actually the deprecated one doesn't have any mobile version. I started working on it presuming it was the same as the others and then i noticed it was not, still i left something "misaligned", i fixed that. 
In general, by the way, we only added a parameter (src_mobile) and renamed src in src_desktop (there is no usage of these templates so it's safe), in EC we don't pass the src_mobile, so there shouldn't be any "wrong" logo in EC in mobile.
The logic is that for EC we need the generic ecl-site-header-core__logo-image class, while in EU we need ecl-site-header-core__logo-image--desktop and ecl-site-header-core__logo-image--mobile, both are set to make everything work in the two systems.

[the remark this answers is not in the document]
planctus: We only added a parameter (src_mobile) and renamed src in src_desktop (there is no usage of these templates so it's safe), in EC we don't pass the src_mobile, so there shouldn't be any "wrong" logo in EC:
```
{% include '@ecl-twig/ec-component-site-header-harmonised/ecl-site-header-harmonised.html.twig' with 
{
logo: {
title: "European Commission",
alt: "European Commission logo",
href: "/example",
language: "en",
src_desktop: "static/media/logo--en.30b933cc.svg"
},
```
I pushed an "update" on the docs, we mention now that the parameter is for EU only. 

For the rest the logic is that for EC we need the generic `ecl-site-header-core__logo-image` class, while in EU we need `ecl-site-header-core__logo-image--desktop` and `ecl-site-header-core__logo-image--mobile`, both are set to make everything work in the two systems.

src/ec/packages/ec-component-site-header-standardised/__snapshots__/site-header-standardised.test.js.snap
planctus: yeah, it was wrong the template, only of the standardised, fixed it ;)

PR #514, opened 2020-07-29
feat(scripts): Creating system aware prettier tasks for php - front-1380

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


package.json
```diff
@@ -16,14 +16,16 @@
...
+    "php:dist": "npm-run-all php:clean render-php-js storybook-build-php:* php:copy php:dist-fix",
     "php:clean": "npm-run-all php-clean:*",
     "php:dist-fix": "./node_modules/@ecl-twig/php-storybook/resources/fix-build.js",
     "php:install": "lerna --scope \"@ecl-twig/php-storybook\" run setup",
     "postinstall": "lerna run postinstall && lerna run prepublish && patch-package",
     "pretty:check": "prettier --check \"src/**/*.{md,mdx,html,json,yml,js}\"",
     "pretty:fix": "prettier --write \"src/**/*.{md,mdx,html,json,yml,js}\"",
-    "pretty:php": "prettier --write \"php/*/!(php_storybook)/**/!(.storybook)/*.{html,js}\"",
+    "pretty:php": "run-p pretty-php:*",
+    "pretty-php:ec": "prettier --write \"php/*/!(eu)!(php_storybook)/**/!(.storybook)/*.{html,js}\"",
```
kalinchernev, line 27: Prettier CLI signature is `prettier [options] [file/dir/glob ...]`.
Seems to me you're trying to use glob patterns. Prettier uses [globby](https://www.npmjs.com/package/globby). Please double-check patterns rules for negation and or statement.

Few remarks:
- having pattern of `foo:bar` and `foo-bar` is not consistent
- `!(eu)!(php_storybook)` => `!(eu|php_storybook)`
- `!(.storybook)` => `!.storybook`
- what you are trying to achieve is split of system-related prettier tasks and make formatting focused on the specific system, so a more performant logic will be to specify system and format selectively with narrowing rather than inverting logic and keeping greedy algorithms

```
prettier --write "php/packages/ec/**/*.{html,js}"
```
Will suffice and do the same for ec system without any negation and will be faster because you specify php/packages/ec which helps narrowing down actual selection of files.

planctus: couldn't agree more, i've changed also the naming of these new scripts. thanks for the review ;)

PR #539, opened 2020-08-26
feat(netlify-cms): Setting netlify-cms on ecl-twig-js - front-1547

planctus:
# PR description

Since we are considering netlify-cms for ECL i wanted to give it a try on ECL-Twig where we also have many md files both as the component documentation and the library documentation.
In ecl-twig it's easier because the content of these md files is simpler, maybe, the value of using a "cms" is clearly limited in this case, but still it offers a nicer experience for free.

The setup is for git-gateway backend, with the netlify identity service sending an invite for the single users (5 maximum in the free plan), the editorial workflow is in place, although the "roles" have not been explored to see how to properly configure it, not sure it will be possible to limit the operations allowed by "role", for instance and potentially from the cms interface the whole flow is under control, including the merge of the pr, corresponding to the "publish" operation.

The files have been statically defined, as far as i understand is not possible to mix up the approaches, with some files statically defined and still the possibility of adding new ones.

At the moment i'm the only developer who was invited, if you are going to review this i will send an invite (it will not work as it is, due to the fact that we don't have the cms in production, basically, so you will have to use the url of the preview for this branch replacing the production url in the link)

Updates:
This contains also a custom widget for the markdown editor in the cms for generating tables, it is a sort if an "hack", the widget shows you two fields for the columns and rows number, it generates a markdown table with "dummy" content which is shown in the preview. To edit the table you switch from rich text to markdown and then back, this way the widget gets replaced by the markdown table.

Styles have been applied to both panels, compiling the ECL files used in the website but using html tags and not classes.
The styles in the editor pane are heavily styled already and only a few of the custom style we define is going to be used (one of this is the table 100% wide which is good)

## QA Checklist

- [ ] I have provided tests


static/admin/config.yml
```diff
@@ -0,0 +1,356 @@
...
+  squash_merges: true
+  commit_messages:
+    create: 'feat({{collection}}): create {{collection}} “{{slug}}”'
+    update: 'feat({{collection}}): update {{collection}} “{{slug}}”'
+    delete: 'feat({{collection}}): delete {{collection}} “{{slug}}”'
+    uploadMedia: 'Upload “{{path}}”'
+    deleteMedia: 'Delete “{{path}}”'
+
+media_folder: static/images
+publish_mode: editorial_workflow
```
kalinchernev, line 14: Unnecessary I think. There won't be too many people doing too many things I think.

static/index.html
```diff
@@ -6,6 +6,23 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <title>ECL Twig</title>
+    <script>
+      if (window.location.hash.match('access_token')) {
```
kalinchernev, line 10: You can surely have the checks in an or statement or `[...list].includes(match)` way of combining the several checks

static/admin/config.yml
```diff
@@ -0,0 +1,356 @@
+backend:
+  name: git-gateway
+  repo: ec-europa/ecl-twig
+  branch: develop
+  squash_merges: true
+  commit_messages:
+    create: 'feat({{collection}}): create {{collection}} “{{slug}}”'
```
kalinchernev, line 7: This pattern looks a bit strange looking at the open pull requests.

kalinchernev (commented):
No issues found reading the code and since it works, i don't see a reason not to move on with the setup.
Having in mind the ever-changing structure of the target team for using the cms, i'd say why not creating 1 generic user (bot) with access to the repo which is shared to be used within the design team?

kalinchernev (dismissed):

static/admin/config.yml
planctus: it looks ok, to me.
`feat(Component docs): update Component docs “ecl_compliance”`
where collection is "Component docs" and "slug" is the item name.

static/admin/config.yml
planctus: Sure, the whole cms thingy is unnecessary here on twig, i'd say. The idea is to use this configuration to test in advance what would be the overall experience of an "editor" of these documents in ECL, including the github workflow, that's why i set it here.

static/index.html
planctus: Yep, we can look for ways to simplify this.

planctus:
I agree that the "users list" that will have access to the ECL cms needs to "defined", to understand the best approach for trying to avoid expenses and still offering an "editorial" workflow with roles, potentially. 
This part is still to be "explored" to come out with a configuration that would serve our needs.
The machine user per se could be a solution, maybe, but it would also mean that there is no distinction between different users, not sure what would happen with multiple sessions, things that we need to consider, let's use this period of time to try different set up on twig, so that we will be prepared for when it comes to ECL.

PR #556, opened 2020-09-08
feat(select): Adding support for disabled and selected options - front-1590

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-select/ecl-select.html.twig
+++ b/src/ec/packages/ec-component-select/ecl-select.html.twig
@@ -144,14 +144,20 @@
     {% endif %}
       {{- _extra_attributes|raw -}}
     >
-      {% for option in _options %}
-        <option
-          value="{{ option.value }}"
-        {% if option.selected is defined and option.selected is not empty %}
-          selected
+      {% for _option in _options %}
+        {% set _option_extra_attributes = '' %}
+        {% if _option.extra_attributes is defined and _option.extra_attributes is not empty %}
+          {% for _option_attr in _option.extra_attributes %}
+            {% if _option_attr.value is defined %}
+              {% set _option_extra_attributes = _option_extra_attributes ~ ' ' ~ _option_attr.name ~ '="' ~ _option_attr.value|e ~ '"' %}
+            {% else %}
+              {% set _option_extra_attributes = _option_extra_attributes ~ ' ' ~ _option_attr.name %}
+            {% endif %}
+          {% endfor %}
         {% endif %}
-        >
-          {{- option.label -}}
+
+        <option value="{{ _option.value }}" {{ _option_extra_attributes|raw }}>
+          {{- _option.label -}}
         </option>
       {% endfor %}
     </select>
--- a/src/ec/packages/ec-component-select/ecl-select.html.twig
+++ b/src/ec/packages/ec-component-select/ecl-select.html.twig
@@ -145,7 +145,12 @@
       {{- _extra_attributes|raw -}}
     >
       {% for option in _options %}
-        <option value="{{ option.value }}">
+        <option
+          value="{{ option.value }}"
+        {% if option.selected is defined and option.selected is not empty %}
+          selected
+        {% endif %}
+        >
           {{- option.label -}}
         </option>
       {% endfor %}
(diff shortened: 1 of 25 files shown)


src/ec/packages/ec-component-select/ecl-select.html.twig
```diff
@@ -144,9 +144,20 @@
     {% endif %}
       {{- _extra_attributes|raw -}}
     >
-      {% for option in _options %}
-        <option value="{{ option.value }}">
-          {{- option.label -}}
+      {% for _option in _options %}
+        {% set _option_extra_attributes = '' %}
+        {% if _option.extra_attributes is defined and _option.extra_attributes is not empty %}
```
kalinchernev, line 149: Having a property `extra_attributes` inside each option is not what I had in mind with adding a way to spread over attributes from the option. The API is not good in this way IMO. The data should be in the same shape as now and possibly have new attributes like `selected` in addition. This is how any system would expect the data structure.

One thing coming tom mind is having a destructuring or spreading, but twig does not support these types of syntax. Found [this](https://github.com/twigphp/Twig/issues/3291#issuecomment-627898820) but not a full solution we'll need to specify attributes again and not just pass them along. `merge` is also not ideal maybe.

I checked [how Drupal select template is doing it](https://api.drupal.org/api/drupal/core%21modules%21system%21templates%21select.html.twig/8.3.x) and it seems to me we won't have the way to make the solution as flexible as I imagined. We'll have to use the `selected` specifically.


```twig
{% spaceless %}
  <select{{ attributes }}>
    {% for option in options %}
      {% if option.type == 'optgroup' %}
        <optgroup label="{{ option.label }}">
          {% for sub_option in option.options %}
            <option value="{{ sub_option.value }}"{{ sub_option.selected ? ' selected="selected"' }}>{{ sub_option.label }}</option>
          {% endfor %}
        </optgroup>
      {% elseif option.type == 'option' %}
        <option value="{{ option.value }}"{{ option.selected ? ' selected="selected"' }}>{{ option.label }}</option>
      {% endif %}
    {% endfor %}
  </select>
{% endspaceless %}
```

Let's use this approach is my suggestion

kalinchernev (changes requested):

src/ec/packages/ec-component-select/ecl-select.html.twig
planctus: This was the initial way i started, the ticket was indeed about the defaultValue, let's say, but options support many more attributes and it would be inconvenient to list them all one by one, the scope was a bit enlarged, then.
This way for instance you would not be able to pass a `disabled` option, plus you don't have any way to pass all the other allowed attributes for the option element, global attributes and event attributes.
https://www.w3schools.com/tags/tag_option.asp

It seems fine to me to define the required properties independently (label and value) and put all the rest in an array the same way we do for the root element of each component. 
This keeps the current structure, just adding an additional key. 

A way in between could be to define `selected` and `disabled` as properties but still providing an `extra_attributes` key where to put additional ones or if we consider the "general" attributes out of scope (why?) at least we should also consider the `disabled` attribute.

What would you prefer, @kalinchernev?

kalinchernev: How about trying to read the properties with the [`attribute`](https://twig.symfony.com/doc/2.x/functions/attribute.html) function [theoretically supported](https://nightlycommit.github.io/twing/language-reference/functions/attribute.html) by twing?
I.e. getting data passed to the option without the `extra_attributes` and setting them via an iterator?

planctus: yeah but the problem is also that you cannot iterate on the object properties in twig unless it's a /Traversable one, anyway, i think here we need to "fix" something, let's stick to the problem, it's enough to fix that. 
For this is enough to include "disabled" and "selected", i believe, we do it "statically" as we already do it for the select itself, easy.
I will change the code this way, but still we can wait for this to be merged, if in ECL we are going to change anything in the specs file to "demo" the default value and the disabled option i will adapt this pr accordingly.

kalinchernev (changes requested):
Maybe we can support both `isDisabled` and `disabled`. I'd suggest the second if we have to pick one.

kalinchernev (approved):

PR #584, opened 2020-11-26
feat(gallery): Add support for external video iframe - FRONT-1817

papegaill:
# PR description

## QA Checklist

- [ ] I have provided tests


src/ec/packages/ec-component-gallery/ecl-gallery-item.html.twig
```diff
@@ -53,6 +55,9 @@
     {# Use Arial only for image. #}
     {% set _use_arial = 'aria-label="' ~ _image_alt ~ '"' %}
   {% endif %}
+  {% if _item.embedded_video is defined and _item.embedded_video is not empty %}
```
planctus, line 58: I find all this block very confusing, it was already, eh, but now a bit more:
basically what we do is to give priority to the embedded_video, then to the video, then to an image object and finally to single properties in the item, right?
Could we restructure the whole logic here so that is a single if (embedded_video), elseif (video), elseif (image), else?
Then, since we are still using the same markup i think the media_href variable can still be used for them all, i see that we need an additional one to handle the custom attribute for the embedded_video, but that could be a different one.
Another confusing thing in the pre-existing logic that we could fix here is that we set the aria_label only for the image object case, but the variable holding its value is defined outside of the logical block for it, i would put it inside, instead.

src/ec/packages/ec-component-gallery/ecl-gallery.html.twig
```diff
@@ -18,7 +18,10 @@
     - "description" (string) (default: '')
     - "meta" (string) (default: '')
     - "icon" (object) (default: {}): object of type icon
-    - "share_path" (string) (default: '')
+    - "image" (optional) (object) (default: {})
```
planctus, line 21: Thanks for updating this ;)

planctus (changes requested):
The test snapshot needs to be updated, for the comment about the logical block i think i would be clearer, making more sense, but up to you to see if you think the same ;)
Thanks!

src/ec/packages/ec-component-gallery/ecl-gallery-item.html.twig
papegaill: Embedded_video act like an image item, with the exception of having iframe url for the href and a specific data attribute, so if I want to make everything in one condition block I have to repeat image logic. Not sure it's better.

_use_arial is defined within the image condition: https://github.com/ec-europa/ecl-twig/blob/8436d4ea429a8f447d580d771b671c5622e89a41/src/ec/packages/ec-component-gallery/ecl-gallery-item.html.twig#L56 not sure to understand this?

papegaill: @planctus snapshots updated ;)

planctus (approved):

PR #587, opened 2020-12-15
feat(raw-filter): Removing usage of the raw filter in all the components - FRONT-1909

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests


emeryro (commented):
I don't know if it is linked to this PR, but on the preview (https://front-1909--ecl-twig-php.netlify.app/ec/?path=/story/components-accordion2--default), all javascript behavior do not work anymore (issue ecl-auto-init, as it seems). 
It works fine on current version (https://ecl-twig-php.netlify.app/ec/?path=/story/components-accordion2--default)

planctus:
Thanks a lot for reviewing this, @emeryro, the problem you saw in the js preview was due to the fact that it was not up to date due to failures in the drone task, that behaviour is indeed a "bug" in twing, that i fixed by setting the autoescape to false in js. In php this is not the case and the preview was also up to date. Thanks to your report i could insist a bit more to understand the reason why the storybook build was failing in drone (unfortunately there is some random failure in that task since a long time) and regenerating the yarn.lock by resetting node_modules fixed this, so now the preview is ok and the js is working also there.

emeryro (approved):

PR #593, opened 2021-01-26
fix(raw-filter): Using raw in safe context to avoid double escaping - FRONT-1940

planctus:
Some explanation:
we blindly followed the request by the platform of removing all the raw filter usages in the library, first, and this lead to some, different misbehaviours:

- In our scenario, all the specs containing html would cause escaping by the twig engine and the html would be rendered as a string, this is not a problem on platform side because with drupal you have ways to mark your html as "safe" for twig.
- Due to this we told twing not to autoescape in 2.35.1 and by doing this we fixed the problem on our side, in the js storybook
- We did not do the same for php and indeed the demo in https://ecl-twig-php.netlify.app shows all the markup in variables as a string. (https://ecl-twig-php.netlify.app/ec/?path=/story/components-timeline--default as an example)
- It's worth mentioning that after a long discussion, drupal 8 enabled the autoescape in twig, which indeed created many issues, but they stick to that, due to the security holes that they were otherwise having.
- There are a few different use cases of html in variables in our templates, one is the aforementioned "specs" property that contains html, but also in many other cases we were storing html in variables and by removing the raw filter we made them appear as strings, as long as the autoescape was in place (https://ecl-twig-php.netlify.app/ec/?path=/story/components-date-block--default).
- In this pr what happens is that we reintroduce the raw filter for attributes, because of the way we build them, indeed, they would be double escaped and this would completely break the js (this is what is in the issue: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1940),  all the strings provided to the template for the attributes in this regard are escaped with the html_attr strategy in the template, so that's why we can then safely use raw when we print the string we built.
- Mainly this is a story about attributes, but there we other cases where the templates were storing markup in variables, this would break when not printing them with raw, so in some other cases code was reorganised in order not to store html in variable (the date block component, for instance).
- Still the autoescape needs to be disabled in our scenario because all the demoes with specs containing html would break but this is risky, because we have to take care, from now on, of not storing html in variables, we would not see the problem generated by this due to the fact that we don't have the twig autoescape in place.
- The test have been run one by one to ensure, before disabling the autoescape, that the only "breaking" things were about data provided by the specs, so i'm fairly confident that the templates are now ok, but if you really want to test this, please remove the `autoescape: false` from the` environment.js` file in the ec storybook folder and run the test, the things breaking should all be related to data provided by the specs.
- As a final consideration, though, i'm confused about what platform asked us in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1909, they showed us a patch where they were using `e('html_attr')` for the attributes but you can see here https://twig.symfony.com/doc/1.x/filters/escape.html that the `html_attr `strategy for escaping has been added in twig 1.9, platform is using, in the most recent version, 1.44, so i don't know...

# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/php/php-storybook/bootstrap.php
+++ b/php/php-storybook/bootstrap.php
@@ -14,5 +14,5 @@ $eu_packages_folder = Path::canonicalize(__DIR__ . '/../../src/eu/packages/');
 $loader = new \Twig\Loader\FilesystemLoader($ec_packages_folder);
 $loader->addPath($eu_packages_folder, 'ecl-twig');
 $loader->addPath($ec_packages_folder, 'ecl-twig');
-$twig = new \Twig\Environment($loader);
+$twig = new \Twig\Environment($loader, ['autoescape' => false]);
 
--- a/src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
+++ b/src/ec/packages/ec-component-accordion/ecl-accordion.html.twig
@@ -40,9 +40,9 @@
 {% if extra_attributes is defined and extra_attributes is not empty and extra_attributes is iterable %}
   {% for attr in extra_attributes %}
     {% if attr.value is defined %}
-      {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value|e('html_attr') ~ '"' %}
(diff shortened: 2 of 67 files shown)


kalinchernev:
I think it might be netlify has changed something on dns level for the preview not to work. You can try changing the pattern.

PR #594, opened 2021-01-27
fix(raw-filter): Using raw in safe context to avoid double escaping - FRONT-1940

planctus:
# PR description

Some explanation:
we blindly followed the request by the platform of removing all the raw filter usages in the library, first, and this lead to some, different misbehaviours:

In our scenario, all the specs containing html would cause escaping by the twig engine and the html would be rendered as a string, this is not a problem on platform side because with drupal you have ways to mark your html as "safe" for twig.
Due to this we told twing not to autoescape in 2.35.1 and by doing this we fixed the problem on our side, in the js storybook
We did not do the same for php and indeed the demo in https://ecl-twig-php.netlify.app shows all the markup in variables as a string. (https://ecl-twig-php.netlify.app/ec/?path=/story/components-timeline--default as an example)
It's worth mentioning that after a long discussion, drupal 8 enabled the autoescape in twig, which indeed created many issues, but they stick to that, due to the security holes that they were otherwise having.
There are a few different use cases of html in variables in our templates, one is the aforementioned "specs" property that contains html, but also in many other cases we were storing html in variables and by removing the raw filter we made them appear as strings, as long as the autoescape was in place (https://ecl-twig-php.netlify.app/ec/?path=/story/components-date-block--default).
In this pr what happens is that we reintroduce the raw filter for attributes, because of the way we build them, indeed, they would be double escaped and this would completely break the js (this is what is in the issue: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1940), all the strings provided to the template for the attributes in this regard are escaped with the html_attr strategy in the template, so that's why we can then safely use raw when we print the string we built.
Mainly this is a story about attributes, but there we other cases where the templates were storing markup in variables, this would break when not printing them with raw, so in some other cases code was reorganised in order not to store html in variable (the date block component, for instance).
Still the autoescape needs to be disabled in our scenario because all the demoes with specs containing html would break but this is risky, because we have to take care, from now on, of not storing html in variables, we would not see the problem generated by this due to the fact that we don't have the twig autoescape in place.
The test have been run one by one to ensure, before disabling the autoescape, that the only "breaking" things were about data provided by the specs, so i'm fairly confident that the templates are now ok, but if you really want to test this, please remove the autoescape: false from the environment.js file in the ec storybook folder and run the test, the things breaking should all be related to data provided by the specs.


emeryro (approved):
it seems good to me.
Let's wait for Plateform's feedback

src/ec/packages/ec-component-date-block/ecl-date-block.html.twig
```diff
@@ -41,32 +41,38 @@
...
+<time class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  {% if _date_time is not empty %}
+    <span class="ecl-u-sr-only">{{ _date_time }}</span>
+  {% endif %}
+  <span class="ecl-date-block__day" aria-hidden="true">{{ _day }}</span>
+{% if (_month is not empty or _month_full is not empty) %}
   {% if _month_full is not empty and _month_full != _month %}
-    {% set _month_markup = '<abbr title="' ~ _month_full ~ '" class="ecl-date-block__month" aria-hidden="true">' ~ _month ~ '</abbr>' %}
+  <abbr
+    title="{{ _month_full|e('html_attr') }}"
```
sergepavle, line 63: These changes reminded me that looks like have to do that in other components. I saw in other components we leave construction like that: `attrname="{{ _variable }}"` with relying on enabled autoescape in twig.
But problem is that autoescape in twig uses HTML strategy (https://twig.symfony.com/doc/1.x/tags/autoescape.html ). So it will be not equal to html_attr.
By following a more strict approach and consistency we have to do the same in other components.
What do you think?

planctus: I agree that if we do these checks we have to do them consistently. I will browse again the library to find other use cases of attribute's values not being escaped, the idea is indeed to have an overall approach and use it consistently, enforcing "security" on attributes using the right escaping strategy from now on.

sergepavle: We had a discussion today with @ademarco and in the context of changes related to security updates, discovered that last changes not all of them fully in line with OWASP recommendations. For instance according to documentation of twig: https://twig.symfony.com/doc/3.x/filters/escape.html and  https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#rule-2---attribute-encode-before-inserting-untrusted-data-into-html-common-attributes we shouldn't use everywhere only html_attr. Looks like for some special attributes we have to use strategy according to the purpose of the attribute.
As I understand for `src` and `href` we have to use `url` strategy and for `style` - `css`.

planctus: makes sense, will spot these specific use cases, thanks for reporting it ;)

planctus: ummm...no, the docs on twig say that `url` strategy should not be used for full urls but only for segments, it escapes the slashes, so even a path would be broken, the hashes too, so that anchor links would be broken... My feeling is that we can stick to `html_attr` as a stronger strategy than the default html, and that's it..

planctus: So...we started having issues on our side as soon as we extended the scope of this pull request, basically to apply a stricter escaping strategy on all the attributes.
Therefore i decided to revert the latest changes and bring back this pull request to its former scope, fixing the double escaping introduced by the removal of the `|raw` filter that was breaking the js and the attribute's functionality in general.

I think this is the safest thing to do, it all originated from the security issues related to the use of `|raw`, we now only use it for strings that we have hardcoded in the templates or that we constructed with parameters that we always previously escaped with `e('html_attr')`. 

All the other use cases of attributes that we set in html in the template and where only their value comes from a variable are untouched, those don't need the usage of `|raw` to avoid double escaping, indeed.

sergepavle: I suppose it is fine.

PR #598, opened 2021-02-16
feat(timeline): Content is optional and title mandatory - FRONT-2036

planctus:
# PR description

## QA Checklist

- [ ] I have provided tests

### DIFF
--- a/src/ec/packages/ec-component-timeline/ecl-timeline.html.twig
+++ b/src/ec/packages/ec-component-timeline/ecl-timeline.html.twig
@@ -9,10 +9,10 @@ Parameters:
     - "to": (integer) (default: items.length) Item index after which to resume displaying timeline items
   - "items" (array) (default: []):
     - "id": (string) (optional) (default: '')
-    - "label": (string) (default: '')
+    - "label": (string) (default: '') To be used as a "timestamp"
     - "title": (string) (default: '')
-    - "content": (string) (optional) (default: '')
-  - "icon_path" (string) (default: ''): file containing the svg icons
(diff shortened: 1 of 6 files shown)


src/ec/packages/ec-component-timeline/ecl-timeline.html.twig
```diff
@@ -8,10 +8,10 @@ Parameters:
     - "from": (integer) (default: items.length) Item index after which to start hiding timeline items
     - "to": (integer) (default: items.length) Item index after which to resume displaying timeline items
   - "items" (array) (default: []):
-    - "id": (string) (default: '')
+    - "id": (string) (optional) (default: '')
     - "label": (string) (default: '')
     - "title": (string) (default: '')
-    - "content": (block) (default: '')
+    - "content": (string) (optional) (default: '')
```
emeryro, line 14: Won't it be an issue to define the content as a string only?
According to the usage page, users can put other type of content in it (media for instance)

planctus: well, i'd say that in these docs on all our templates we refer to the data type, whatever this variable will contain it's going to be used as a string, not an array, not an object, not a boolean, in other words. But you are right that maybe we should treat this, as the other variables that are "allowed" or supposed to contain html (the ones that we were rendering with `|raw`) in a different way, because `string` in those cases might be confused with "plain text", we could say something like `html markup marked as a safe string` to somehow specify that it would be otherwise escaped in an environment with the twig autoescape on.

emeryro (approved):

